PR #1, opened 2017-06-24
Replace corpusc import with skycoin

gz-c:
Do not merge.  Pulls in too many dependencies due to import of skycoin/cipher, which imports skycoin/util, which imports many other dependencies.  Will fix skycoin/util issue


gz-c:
see #2

PR #30, opened 2018-01-04
add docker-compose.yml

djskinner:
Using the recently added Dockerfile and instructions I've added a `docker-compose.yml` file.

This condenses the series of commands in the README into an easy to read config and allows spinning up a fully functional skywire node and manager with a single line when using docker-compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,3 +116,11 @@ docker run -ti --rm \
       -address :5000 \
       -web-port :6001
 ```
+
+### Docker Compose
+
+```
+docker-compose up
+```
+
+Open [http://localhost:8000](http://localhost:8000).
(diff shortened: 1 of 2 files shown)


gz-c:
@etiennetremel is this ok?  should we leave the original instructions in the readme still, or is it ok to expect people to have docker-compose?

etiennetremel:
That's a good idea to add the docker-compose with both the manager and node there, but I would suggest keeping the basic Docker instruction. Docker-compose is just a tool to connect things together and should be used as a base if you want to integrate with other scheduler like Kubernetes, Mesos, etc.
@djskinner do you mind reverting and maybe add a sub-section for the docker-compose instructions?

djskinner:
No problem, will do.

PR #44, opened 2018-03-27
Fix docker build issues

### LINKED ISSUE #43: Multiple Docker build issues
While working on #42 I detected several issues:
-  vendor/github.com/mattn/go-sqlite3 is outdated and does not compile. It throws the following error
`# github.com/skycoin/skywire/vendor/github.com/mattn/go-sqlite3
vendor/github.com/mattn/go-sqlite3/sqlite3_go18.go:18:10: undefined: SQLiteConn`
- `node` throws the following error
`ERRO[0064] after launch error: fork/exec ./sockss: no such file or directory`
- There is no image for ARM arch in Docker Hub
I'll be working on this issues ASAP

stdevMauricio:
This solves the first two issues I described in #43.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,12 +3,14 @@
 ARG IMAGE_FROM=alpine:3.7
 FROM golang:1.9-alpine AS build-go
 ARG ARCH=amd64
+ARG GOARM
 
 COPY . $GOPATH/src/github.com/skycoin/skywire
 
 RUN cd $GOPATH/src/github.com/skycoin/skywire && \
-    GOARCH=$ARCH CGO_ENABLED=0 GOOS=linux go install -a -installsuffix cgo ./... && \
+    GOARCH=$ARCH GOARM=$GOARM CGO_ENABLED=0 GOOS=linux go install -a -installsuffix cgo ./... && \
     sh -c "if test -d $GOPATH/bin/linux_arm ; then mv $GOPATH/bin/linux_arm/* $GOPATH/bin/; fi"
+    sh -c "if test -d $GOPATH/bin/linux_arm64 ; then mv $GOPATH/bin/linux_arm64/* $GOPATH/bin/; fi"
 
 
 # skywire manager assets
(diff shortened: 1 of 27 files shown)


olemis:
What is this PR about ? @stdevMauricio could u pls edit PR summary title ? Thnx in advance

stdevMauricio:
This PR updates the sqlite3 dependency, because it didn't compile with the previous source.
I changed the sockss execution path. It started with "./" so it didn't used the $PATH variable. If the binary was moved to another location, as it happens while building the Docker image, it couldn't be found.

vyloy (changes requested):
this change requires sockss in $PATH, which is not every user wants

vyloy (changes requested):
update Gopkg.lock when you update vendor files

gz-c:
@olemis @stdevMauricio what is the status of this?

cypheron:
This is an important fix (also fixes #57)! Executing any command with './' is bad programming because it requires the executing program to be in the executable directory.
@vyloy A user not under control of his $PATH should probably not use linux at all. Not a reason to keep a bad design.
@stdevMauricio I would have created two PR's with each PR addressing one issue. Easier to see for a third person.

cypheron (approved):

stdevMauricio:
I'd like to point out that I followed @vyloy instructions and left the source code untouched. I had to copy the command to the working path for this to work, so it's twice in the file system. I know using './" is bad programming, but I was told to leave it that way. I still think we should remove "./" from the source code. Those are my two cents.

cypheron:
@stdevMauricio No one likes relative paths, they only cause pain. I would create a new PR with just the changes regarding execution paths. It's not only the ./sockss command, there are many more in api.go. I would suggest something similar to this style `exec.Command(filepath.Join(gopath, fmt.Sprintf("%s%s", scriptPath, "unix/check")))` so basically joining gopath with the executable name, making the paths absolute and much more reliable. With that change it is no longer needed to tell people to `cd` into $GOPATH in the guides, which is embarrassing btw, as commands will be able to be run from any CWD... Please change this soon.
Update: #74 addresses the path issue as a separate concern.

gz-c:
The setup instructions should tell the user to add $GOPATH/bin to $PATH via bashrc/zshrc.  However I don't know how this would affect windows users

PR #46, opened 2018-04-08
Update the static files

redsh4de:
Changed slight grammar and spelling errors in the static files


gz-c:
Don't modify the static files, change the source, the project owner will update the static files afterwards

PR #93, opened 2018-07-13
Ui

IvanPayan:
Implemented design on Login page in Skywire manager project.


mahansky (changes requested):
When you change `styleUrls: ['./login.component.css']` to `styleUrls: ['./login.component.scss']` and rename the file as well, you can use scss syntax in it, which is recommended. Saves a lot of repetition (in this case, you can nest everything under `.login-container { ... }`

There is an outline on the login button which looks bad since it does not render border-radius. I suggest removing the outline.

The input needs a placeholder (and a smaller font-size so it doesn't stand out so much)

IvanPayan:
I has made the changes you have commented on the last commit

mahansky (approved):

PR #94, opened 2018-07-13
Node status and running apps

mahansky:


IvanPayan:
I'm trying to access at page http://localhost:4200/nodes/{{key}}, and the result is 404 page not found

mahansky:
Router uses hash notation so the url is localhost:4200/#/nodes/key

IvanPayan (approved):

PR #97, opened 2018-07-17
Node list view implemented designs

IvanPayan:


therealssj:
think you could attach some screenshots of how it looks?
thanks for your work! :D

mahansky (approved):

mahansky:
@IvanPayan I think it would be better to use Material Icons instead of adding custom SVGs. Material Icons are already installed in the project.

IvanPayan:
captura de pantalla 2018-07-18 a les 15 35 46

@therealssj this is a screenshot of the node list view :)

IvanPayan:
@mahansky okey! I'll use the Material Icons. For those icons that do not exist in the Material Icons collection, shall we use the SVG or change it to a Material Icon that looks similar?

mahansky:
Similar material icon

PR #109, opened 2018-08-01
Terminal

mahansky:
- Updated typescript (and angular) to support TextDecoder
- Created Terminal component using xterm.js


IvanPayan:
@mahansky the "Use the terminal carefully" dialog is not showing before the terminal is not shown. It should be opened right after clicking the "Terminal" button. Then when the user selects "OK" the terminal will be shown (you just have to call "terminal" function instead of "openTerminal", in the actions.component file.

captura de pantalla 2018-08-02 a les 10 09 26

Issue #111, opened 2018-08-02
Fix Docker Cloud build for ARM


olemis:
New base image needed .

olemis:
see #193 for #191

PR #112, opened 2018-08-02
[docker] refs #111 - Disable builds for ARM v5 and v7

olemis:
Temporary solution for #111 

Changes :

- Enable the builds for versions supported by `alpine` image
- Disable the rest

### DIFF
--- a/hooks/build
+++ b/hooks/build
@@ -2,7 +2,7 @@
 docker build -f $DOCKERFILE_PATH -t $IMAGE_NAME .
 
 # ARM builds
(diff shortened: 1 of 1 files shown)


olemis:
@stdevMauricio pls take a look into this issue

olemis:
This is issue has been solved . Using `busybox` as base image now .

some4:
current U-BOOT will not compile with ARM v6 and below

olemis:
> current U-BOOT will not compile with ARM v6 and below

@some4 if that's the case then create another issue to work on it . This PR was aimed at fixing  #111 . Since the problem was caused by using `alpine` as base image and we are not using it now , no further work needed in here .

some4:
@olemis Oh I meant to note that it's incompatible--v6 and below won't work with upstream there's nothing we can do about it

PR #137, opened 2018-08-31
Whitelisting link update

asxtree:
Updated the whitelisting link that was pointing to github skycoin whitelist


asxtree:
Pull only the rules.md where I modified the whitelisting link

asxtree:
Pull only the rules.md where I modified the whitelisting link

PR #140, opened 2018-09-06
Rewards size updated

asxtree:
Updated the reward sizes per node of DIY and officials.


gz-c:
Rewards should probably be dated since they vary per month

asxtree:
Opened a new one

Issue #90, opened 2018-07-10
Node app/SSHS app disappear from UI after Skywire web process "manager" restarts


asxtree:
Hi,

I've encountered this Web UI bug which happens when the Skywire web process "skywire-manager" restarts.

**The Behavior** is that the node app or sshs app from all nodes seen in the manager Web UI simply disappears if the `manager` process restarts or is stopped and started again, although if i check via cli on the nodes, both processes (`sockss` and `sshs`) are running. 

**The Workaround** is to restart the `node` process.

!!!!!See the attached photo!!!!

![image](https://user-images.githubusercontent.com/34941232/42532009-2b093030-848e-11e8-99e1-dc62d51d4c9f.png)

mrhodlr:
I have seen this many times as well, but there is more going on than just the Node App disappearing. When you reboot the manager, all nodes seem to lose part of their connection to manager, They still display but all the buttons in the UI become non-responsive.  Terminal window will not open, you cannot click on Settings to check or update Discovery Address, if you click on the Reboot button you get the error message "Manager Token Null"

some4:
I've been trying to isolate this from the system side and would like to add an observation.

Skywire and GOPATH/GOBIN in `/home/$user/`:

If I run Node and Manager as:

- $user: the `.skywire` directory is created in `/home/$user/`

- root: the `.skywire` directory is created in `/home/$user/go/bin/`

What led me to this is I've been following: https://github.com/skycoin/skywire/wiki/Automated-restart-of-node-app to try and lock down the manager app key but I cannot find the "ss" directory e.g.
`/usr/local/skywire/go/bin/.skywire/ss`
or
`~/.skywire/ss`

In both scenarios the permission of `.skywire` created is 0700.

I'm using this as a starting point to find where in the Manager the folder creation happens

Issue #55, opened 2018-05-13
Skywire Manager Glitching out:


ADepic:
I connect to raspberry pi using ssh and enter these two commands:

nohup ./manager -web-dir ${GOPATH}/src/github.com/skycoin/skywire/static/skywire-manager > /dev/null 2>&1 & echo $! > manager.pid

nohup ./node -connect-manager -manager-address 192.168.1.1:5998 -manager-web 192.186.1.1:8000 > /dev/null 2>&1 & echo $! > node.pid

And then launch manager at the local ip address for the raspberry pi.

However when I click on the option of "Check Update", "Settings" , "Connect to node" or "Connect to node > Search Services", the pop-up that appears just dissapears.

With Terminal, I can get a glimpse of a error coming up:
![bad_handshake](https://user-images.githubusercontent.com/24532838/39971394-c7d1a852-56e9-11e8-985c-243a25f918f9.png)

And with reboot, the warning sign comes up, I press "Yes", and then: 
"Error
manager token is null"

I suddenly got this problem after a reboot. I was messing around with some sh scripts.

nikogabs:
Hello
After, 13 hours online, i had the same issue whith all my nodes, impossible to contact them via the manager terminal and i got exactly the same error as above. Idem for the reboot button.
I reboot all my nodes and everything is back up.

PR #148, opened 2018-09-15
Fix disappearing node

jdpwebb:
This should fix the issue where nodes disappear from the manager when the manager is restarted by itself. The modified code updates the manager token in the NodeApi every time the node makes a new connection with a manager.
Addresses issues #90 and #55.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -160,13 +160,14 @@ func (n *Node) connectDiscovery(addr string) (err error) {
 	return
 }
 
-func (n *Node) ConnectManager(managerAddr string) (err error) {
+func (n *Node) ConnectManager(managerAddr string, onConnection func()) (err error) {
 	err = n.manager.ConnectWithConfig(managerAddr, &factory.ConnConfig{
 		Context:       map[string]string{"node-api": n.webPort},
 		Reconnect:     true,
 		ReconnectWait: 10 * time.Second,
 		OnConnected: func(connection *factory.Connection) {
 			go func() {
+				go onConnection()
 				for {
 					select {
 					case m, ok := <-connection.GetChanIn():
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -160,14 +160,20 @@ func (n *Node) connectDiscovery(addr string) (err error) {
 	return
 }
 
-func (n *Node) ConnectManager(managerAddr string, onConnection func()) (err error) {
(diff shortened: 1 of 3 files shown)


pkg/node/node.go
```diff
@@ -160,13 +160,14 @@ func (n *Node) connectDiscovery(addr string) (err error) {
...
 
-func (n *Node) ConnectManager(managerAddr string) (err error) {
+func (n *Node) ConnectManager(managerAddr string, onConnection func()) (err error) {
 	err = n.manager.ConnectWithConfig(managerAddr, &factory.ConnConfig{
 		Context:       map[string]string{"node-api": n.webPort},
 		Reconnect:     true,
 		ReconnectWait: 10 * time.Second,
 		OnConnected: func(connection *factory.Connection) {
 			go func() {
+				go onConnection()
```
gz-c: Does this goroutine leak?

jdpwebb: Good catch. Most of the time, it will not leak. However, it will leak if the node is successful in establishing a connection with the manager, but cannot get the token for some reason. If ```token, err := ioutil.ReadAll(resp.Body)``` *always* returns an error, then the goroutine will never finish, causing a leak.

Some options to fix this:

1. Only try to get a token to connect to the manager for a fixed number of times. For example, only try 60 times (5 minutes at current checking rate), and if the token cannot be successfully received after that time, just exit.
1. Add a channel so the gouroutine can be stopped. If the connection to the manager is closed, then stop the goroutine.

Both methods above would avoid a leak, but I think the first one is simpler and will solve the problem. If the node can connect to the manager, but the manager API is not working, it is unlikely it will suddenly start working after a long time. 

What do you suggest?

gz-c: Do the simpler method. The 2nd method is ideal but the application usually needs to be written from the ground up to support it (e.g. with a `context.Context` everywhere)

gz-c: Unless this method would get called many times a loop, leading to goroutine exhaustion while the stragglers were still waiting to time out

jdpwebb: I think it would be unlikely for this method get called many times because that would mean that the connection to the manager is reestablished many times before any of the goroutines time out. But I agree that it is possible and not the best design.

I looked at the code again and came up with a way for the goroutine to exit if the connection to the manager is closed. I believe this should avoid any leaking. The behavior now is for the node to try and get a token from the manager until the connection is closed, or it is successful.

Issue #100, opened 2018-07-20
Logic error in verifyLogin function in Monitor.go (Low Priority)


BigOokie:
The `verifyLogin` function within  `Monitor.go` appears to have a logic error whereby a session value is read into the variable `hash` however the `pass` variable is checked determine if it is `nil`. Suspect this should be a check of the `hash` variable instead.  

Sections of the associated code below - the variable `pass` is checked twice.

```
verifyLogin(w http.ResponseWriter, r *http.Request, checkDefaultPass bool) bool {
	... 
	pass := sess.Get("user")
	if pass == nil {...}

	hash := sess.Get("pass")
	if pass == nil {...}
	...	
}
```

BigOokie:
Also note that the Manager Web UI appears to enforce its own length restrictions on the user password (must be between 4 and 20 characters)

gz-c:
Fixed by #163

Issue #99, opened 2018-07-20
Manager password restricted to maximum 20 characters (Low Priority)


BigOokie:
**LOW PRIORITY**
The Manager currently restricts the password to 20 characters currently. Given the password is hashed using BCRYPT, the maximum size of the password doesn't really matter as the resulting hash value will be the same size regardless of the length of the password.  Minimum password length check should continue to be enforced.

The following functions in `Monitor.go` contain these restrictions: `UpdatePass` and `Login`. Could reduce code by removing these as unsure they add value.

```
func (m *Monitor) Login(w http.ResponseWriter, r *http.Request) (result []byte, err error, code int) {
	...
	if len(pass) < 4 || len(pass) > 20 {...}

	err = checkPass(pass)
	...
}

func (m *Monitor) UpdatePass(w http.ResponseWriter, r *http.Request) (result []byte, err error, code int) {
	...
	if len(oldPass) < 4 || len(oldPass) > 20 {...}
	...
	if len(newPass) < 4 || len(newPass) > 20 {...}
	...
}
```

gz-c:
TODO:

- Remove password restrictions (is a minimum even necessary?)
- Make sure the bcrypted password is also salted

BigOokie:
> Remove password restrictions (is a minimum even necessary?)

Not wanting to complicate things - the default password is currently `1234`. On that basis, why not simply enforce a 4 character minimum. Technically I assume you would check that the password is at least 1 character (even if not enforcing password length) as a test to ensure there is a password or not.

gz-c:
Yes would have to be at least 1 character.  I don't think it matters to us if a user wants to use a 1 character password

PR #163, opened 2018-10-16
Issue #100. Address logic error in verifyLogin function in Monitor.go

BigOokie:
Changes to `monitor.go`:
- `UpdatePass` function:
  - Remove password length check restrictions. Previously the password would be rejected if its length was < 4 or > 20. Now there is no maximum restriction on the length of the password, and the minimum length is > 0 (i.e. 1 character).
  - Updated the error messages returned by the function.
  - Use new `isDefaultPass` (defined in utils.go) to check if the users password is the default ("1234")
  - Use newly renamed `userHasDefaultPass` function to determine if the user has the default password. This function was previously named `isDefaultPass`

- `verifyLogin` function:
  - Correct logic error. Swap use of `pass` with `hash` variable.

Changes to `utils.go`:
- Add comments to each function.
- Renamed `isDefaultPass` function to `userHasDefaultPass`.
- Added new function `isDefaultPass` to check newly provided user plaintext password against the default password "1234"

Was unable to build successfully within my personal fork (without using `dep` to rebuild all project dependencies - which seemed to work, but unsure of the consequences). Performed testing against a local branch of the official `skywire` repo. Tests conducted via the Manager Web GUI.

Please let me know if there are any change that should be made and/or additional testing needed.

### DIFF
--- a/pkg/net/skycoin-messenger/monitor/monitor.go
+++ b/pkg/net/skycoin-messenger/monitor/monitor.go
@@ -629,7 +629,7 @@ func (m *Monitor) UpdatePass(w http.ResponseWriter, r *http.Request) (result []b
 		result = []byte("A new password must be provided.")
 		return
 	}
-	if isDefaultPassCleartext(newPass) {
+	if isDefaultPass(newPass) {
 		result = []byte("The default password cannot be used.")
 		return
 	}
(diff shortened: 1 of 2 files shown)


BigOokie:
This PR addresses issue #99  also

PR #168, opened 2018-11-08
Update testnet_rules.md

Asgaror:
Recent issues demanded to explicitly state that two miners running in the same location are not allowed


Asgaror:
Updated, included new hardware:
- odroid c2
- x96 mini 
- nano pi m4
- developer board IV

PR #174, opened 2018-11-27
Update README.md of newly pushed  one_time-upgrade script 

Asgaror:
- Fix wrong execution command of ./one_time_upgrade (- instead of underscore)
- Change instructions and remove root:~# since people were copying the lines which led to 'command not found' errors.
- Removed link to networking guide bc the terminal in the web interface cannot be used for the upgrade
- Added note about root password change after upgrading since the script sets the root password back to *skywire*
- Added instructions in a new 'Troubleshooting' section, how to delete & regenerate SSH key pairs in case SSH connection issues occur 
- Added fix for *Error: clone operation failed* (git not installed) to Troubleshooting section
- Added warning about relocation of the .skywire folder from /usr/local/skywire/go/bin/.skywire/ to /root/.skywire


stdevPavelmc:
In view of the problems detected with the users, I find it very convenient to apply these changes.

Issue #195, opened 2019-02-20
Feature: Docker multi node scaling support


BigOokie:
Have been working recently with @alexskyfleet on ability to scale multiple instances of Skywire Node within a Docker environment.

I have created a Docker Compose configuration file which supports this by defining both the Manager and the Node as Services, and then allowing Docker Compose to scale the number of nodes.

PR #196, opened 2019-02-20
Docker compose to support multi node scaling support

BigOokie:
Created a new folder `docker` off the root folder to house Docker scripts.
Create a new folder `multi-node` to address #195 
Create a new `docker-compose.yml` containing services defined for both the Manager and Node, and a `Readme.md` providing usage instructions for `docker-compose` and scaling the number of Nodes.

It's possible to substitute the Docker image used within the `docker-compose.yml` file should the default one not be appropriate for you environment.

There appear to be some issue with the use of the `$HOSTNAME` variable in the `docker-compose.yml` on some systems.


BigOokie:
@olemis any feedback on what i've done here would be awesome - I believe your mastery of Docker may be beyond mine, but think whats here may be a good start.

docker/multi-node/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+  swmanager:
+    image: skycoin/skywire
+    ports:
+      - "8000:8000"
+      - "5998:5998"
+  swnode:
+    image: skycoin/skywire
+    #volumes:
+    #  - skywire-data:/root/.skywire
+    command: node -connect-manager -manager-address swmanager:5998 -manager-web swmanager:8000 -address $HOSTNAME:5000 -web-port $HOSTNAME:6001 -discovery-address discovery.skycoin.net:5999-034b1cd4ebad163e457fb805b3ba43779958bba49f2c5e1e8b062482904bacdb68
```
olemis: the discovery address , shouldn't it be parameterized ?

BigOokie: Ok. Didn't think it needed to be, but this was a copy and paste from the docker-compose.yml thats in the root folder. We should possibly look to resolve this also?

BigOokie:
@olemis also note that I also disabled the storage as this was causing a problem. It was being shared among all instances of the Node which didn't work when you scaled - as each would use the same configuration and keys, which means you are scaling identical clones... we want to scale individual instances. I assume there is some way to persist individual storage for each instance? More research needed.

Also the docker image that is in use here contains the old Manager UI. Has a new version of the docker image been built that contains the new UI? I would prefer to use that one if possible.

olemis:
@stdevPavelmc could you pls take a look ?

BigOokie:
> @stdevPavelmc could you pls take a look ?

This is not high priority. Please don't let it distract other work.

stdevPavelmc:
Hi @BigOokie, I'm replicating the environment here to take a look at this. Cheers.

stdevPavelmc:
> > @stdevPavelmc could you pls take a look ?
> 
> This is not high priority. Please don't let it distract other work.

Ok, no problem, I had to refresh the docker environment anyway...

PR #203, opened 2019-03-05
updated cli

### LINKED ISSUE #200: CLI improvements
Refactor to variables instead of structures. Fix problem with http services and `--addr` flag:

Example: 
image

Here `--addr` is a flag of `transport-discovery`, but after using `edge` subcommand cobra takes it as if the flag were to `edge` command, and thus failing.

ivcosla:
Closes #200.

Commands, variables used as flags and auxiliar functions have been put inside `make[name]Command` functions in order to not clash in the scope with others of similar naming.


evanlinjin (approved):
@ivcosla I made various changes to make CLI conform to specs, cobra standards and http://docopt.org/ (for CLI help).

PR #208, opened 2019-03-06
Added multiple paths for config.

### LINKED ISSUE #205: Change behaviour of how skywire-node finds config file.
Should check for config file in this order:
1. CLI argument (if specified, directly use this. Return error if not valid).
2. Environment variable (env prefix for Skywire should be `SW`. If set and not valid, return error).
3. Uses `~/.skycoin/skywire/config.json` if exists.
4. Uses `/usr/local/skycoin/skywire/config.json` if exists.

Similar behavior should happen for apps, app-configs and database files too. This can be defined later on.

ivcosla:
Closes #205 

Changes only for `config.json`. So far `boltdb` path and `app` path are both specified in the config, if we want to take a different approach for them I suggest to do add flags in the cli, and give priority to what is specified in the config file, if they are not specified in the config file then check flag, env, etc.


evanlinjin (changes requested):
Great work! 

Please update README to provide instructions to install into `$HOME`. 

Also update `skywire-cli config` to have "operation modes" (where the default behavior is to make config to run skywire in $HOME and also generate the actual config in $HOME). Use the behavior of `skycoin/skywire-updater` as reference.

evanlinjin (changes requested):
Looks great! However, the Makefile will need to be updated (looking at the changes made in `mainnet` branch).

evanlinjin (approved):

evanlinjin:
@ivcosla review and merge when this passes. Good job with this PR!

PR #237, opened 2019-03-20
Feature/create auto tests for chat application-228

ayuryshev:
Started test-suite in `test/suite`  for chat-application


ayuryshev:
moved to skywire-services

Issue #241, opened 2019-03-23
Write TestingProposals.DRAFT.md


ayuryshev:
Write a document describing:

- the current status of our testing: What tests we are running, how they are running, where they are running
- long term  proposals for the testing: What tests, how and where. 
- create a small „roadmap“: What is most urgent? What can be done later?

jdknives:
Proposal has been created and is being merged into specs.

PR #242, opened 2019-03-25
[WIP] TestingProposals-241

ayuryshev:
I'm publishing right now results to review and discuss

1. TestingProposals.md - the doc which I started, personally I think
that second doc is better for us - it's more specific for skywire
2. PeriodicTableOfTests.md - it's more concentrated on measurement and
decision support
3. periodic-table-of-tests.zsh - is a POC, how to create measurements. It's only for illustration purposes

Closes: #241

### DIFF
--- a/PeriodicTableOfTests.md
+++ b/PeriodicTableOfTests.md
@@ -1,0 +1,200 @@
+# "Periodic Table of Tests"
+
+## Description
+
+### 1. Tests as `Test*`-functions
+
+Description:
+
+    Test as golang `Test*`-function runnable by `go test`
+
+Motivation:
+
+    Part of development process.
+    Why/when needed: to ensure that written code works in principle.
+
+Measurement:
+
+    a. with `go test`:  `go test ./[path]/... -list Test |rg Test --count`
+    b. naive `rg`: `rg Test ./[path]/**/*.go --json |jq -r ".data.stats.matches" |tail -1`
+
+    Methods can/should give different results.
+    Because "naive rg" returns commented tests too.
+    It should/could/must be a point of interest if they differ.
+
+Coverage:
+
+    Using `go test -cover`
+
+### 2. Examples
+
+Description:
+
+    Test as golang  `Example*`-function runnable by `go test`
+    Subtype of golang tests. Combines features of test and documentation.
+
+Motivation:
+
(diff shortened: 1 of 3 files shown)


TestingProposals.md
```diff
@@ -0,0 +1,355 @@
...
+
+## 1.  Tests in  development environment and automated tests
+
+### Description
+
+**Developer environment**:
+
+- very simplistic but allows fast refresh-restart
+- it could be 2-3 nodes  in one-two networks
+- easily integrated with IDEs, etc
```
evanlinjin, line 68: Not priority.

TestingProposals.md
```diff
@@ -0,0 +1,355 @@
...
+-------------------------------------------------------------
+
+## 1.  Tests in  development environment and automated tests
+
+### Description
+
+**Developer environment**:
+
+- very simplistic but allows fast refresh-restart
+- it could be 2-3 nodes  in one-two networks
```
evanlinjin, line 67: We need to be able to easily configure the number of nodes/services for the local network.

TestingProposals.md
```diff
@@ -0,0 +1,355 @@
...
+- absolute logical consistency
+- full coverage of the theme of software testing
+
+-------------------------------------------------------------
+
+## 1.  Tests in  development environment and automated tests
+
+### Description
+
+**Developer environment**:
```
evanlinjin, line 64: There should be two types of developer environments:
1. Local environment (all services and the entire Skywire network runs locally).
2. Public environment (all services except `skywire-node` runs on public servers).

periodic-table-of-tests.zsh
```diff
@@ -0,0 +1,32 @@
+#!/bin/zsh
```
evanlinjin, line 1: Not needed.

PeriodicTableOfTests.md
```diff
@@ -0,0 +1,200 @@
+# "Periodic Table of Tests"
```
evanlinjin, line 1: Not needed.

evanlinjin (changes requested):
Everything that is not commented on by me, I completely agree with.

jdknives (changes requested):
Needs to be merged into skywire-specs rather.

Issue #248, opened 2019-03-27
Apps create multiple transports


ivcosla:
This behavior happens with every app, upon some time of using them with two nodes that have a single transport between them, additional ones will be created by the app.

This can be observed after sending many messages with `chat`, or after using `ssh` for a while, which I advice in order to reproduce this bug, since it appears faster.

It is desired to use and keep a single transport between them.

In order to reproduce the bug:

1) Start two skywire nodes locally, for the second node configuration file, provide different routes for parameter `routing.table.location`, and different ports for chat app (add this to chat args: `"args": ["--addr", ":8001"]`). Bind a different port in `interfaces.rpc`.

If you wish to use ssh app, add `therealssh-client` to app list and follow the rest of the instructions here: https://github.com/skycoin/skywire/tree/mainnet/cmd/apps/therealssh

2) Use `skywire-cli add-transport other-node-pk` to add a new transport.

3) Check that there is a single transport between them with `skywire-cli list-transports`

4) Use apps between the two nodes, after a while call `skywire-cli list-transports` and you should be able to see that multiple new transports between the two nodes have appeared

PR #252, opened 2019-03-28
Deterministic properties of Transports. Linked to "Apps create multiple transports-248"

ayuryshev:
Changes:

1. pkg/transport/manager.go: instead of empy uuid, now it uses namebased uuid based on edges+type+public properties of transport.  This prevents creation of double transports
2. Edges are sorted - so their order is always the same
3. SignedEntry.Signatures are stored in the same order as edges
4. Local() and Remote() now depends on Manager. Transport itself does not have local and remote sides - it's between them

Minor changes: 

1. Makefile: temporarily switched off data races detection for pkg/messaging while  #253  is not solved. Compatibility with last changes in skywire-cli.
2. Updated existing tests: compatibility and refactoring. Coverage became  slightly better
3. Merged several times. Formatted, linted

Test:

Run in separate shells:
NodeA: `make docker-run`
NodeB: `./skywire-cli config && make run`
Third shell: 
```
$  ./skywire-cli add-transport $SW_NODE_B_PK   # repeat this several times
$ ./skywire-cli list-transports
```
It must return only one row. (In current branch it creates duplicate transports)

Note:

To work this PR needs to have renewed setup node. 
With this version of pkg/transport to work.

Closes: #248


pkg/transport/manager.go
```diff
@@ -198,7 +198,8 @@ func (tm *Manager) Serve(ctx context.Context) error {
 
 // CreateTransport begins to attempt to establish transports to the given 'remote' node.
 func (tm *Manager) CreateTransport(ctx context.Context, remote cipher.PubKey, tpType string, public bool) (*ManagedTransport, error) {
-	return tm.createTransport(ctx, remote, tpType, uuid.UUID{}, public)
+	uid := uuid.NewSHA1(uuid.UUID{}, []byte(remote.Hex()))
```
evanlinjin: This implementation is problematic because Transport IDs are used **globally** to identify the transport and a transport is a bi-directional communication between two nodes.

Refer to specs: https://github.com/watercompany/skywire-specs/blob/master/Specifications.md#transport

evanlinjin (changes requested):
We need to give this problem some more thought.

pkg/transport/manager.go
evanlinjin: if we combine the local and remote public keys, where we order the keys based on some factor... That might work.

ayuryshev: Done

pkg/transport/manager.go
```diff
@@ -198,7 +198,8 @@ func (tm *Manager) Serve(ctx context.Context) error {
 
 // CreateTransport begins to attempt to establish transports to the given 'remote' node.
 func (tm *Manager) CreateTransport(ctx context.Context, remote cipher.PubKey, tpType string, public bool) (*ManagedTransport, error) {
-	return tm.createTransport(ctx, remote, tpType, uuid.UUID{}, public)
+	uid := uuid.NewSHA1(uuid.UUID{}, append(tm.config.PubKey[:], remote[:]...))
```
evanlinjin: If the remote is initiating the transport creation, the public keys will be the other way around. We should have an algorithm that puts one public key before another (some way of sorting).

evanlinjin (changes requested):

pkg/transport/manager.go
ivcosla: I think that may work, just having both PKs always in the same order, no matter who is initiating, but yes, we need to consider it carefully.

ayuryshev: Confirmed:
```
 ./skywire-cli add-transport $SW_NODE_B_PK  
 ./skywire-cli --rpc $SW_NODE_B:3435 add-transport $SW_NODE_A_PK
```
Creates two records

Makefile
```diff
@@ -49,8 +49,22 @@ vendorcheck:  ## Run vendorcheck
 	#GO111MODULE=off vendorcheck ./cmd/therealssh-cli/... 	
 
 test: ## Run tests for net
+	-go clean -testcache
```
evanlinjin, line 60: Is this an extra `-` prepended? haha

ayuryshev: `-` is for continuing in case of errors

pkg/transport/manager.go
```diff
@@ -196,9 +213,31 @@ func (tm *Manager) Serve(ctx context.Context) error {
 	return nil
 }
 
+// GetTransportUUID generates uuid.UUID from pair of keys.
+// Generated uuid is:
+// - always the same for a given pair
+// - GenTransportUUID(keyA,keyB) == GenTransportUUID(keyB, keyA)
+func GetTransportUUID(keyA, keyB cipher.PubKey, tpType string) uuid.UUID {
```
evanlinjin: Rename to `MakeTransportID`.

pkg/transport/entry.go
```diff
@@ -16,7 +16,7 @@ type Entry struct {
 	ID uuid.UUID `json:"t_id"`
 
 	// Edges contains the public keys of the Transport's edge nodes (should only have 2 edges and the least-significant edge should come first).
-	Edges [2]cipher.PubKey `json:"edges"`
+	EdgesKeys [2]cipher.PubKey `json:"edges"`
```
evanlinjin: Rename to `EdgeKeys`.

evanlinjin (commented):
Good work so far. Just some tweaks to naming convention.

pkg/transport/entry.go
```diff
@@ -70,6 +95,33 @@ type SignedEntry struct {
...
+// Index returns position of a given pk in edges
+func (se *SignedEntry) Index(pk cipher.PubKey) byte {
+	if pk == se.Entry.Edges()[1] {
+		return 1
+	}
+	return 0
+}
+
+// SetSignature sets Signature for a given PubKey in correct position
+func (se *SignedEntry) SetSignature(pk cipher.PubKey, secKey cipher.SecKey) {
```
evanlinjin: Rename to `Sign`.

pkg/transport/entry.go
```diff
@@ -70,6 +95,33 @@ type SignedEntry struct {
 	Registered int64         `json:"registered,omitempty"`
 }
 
+// Index returns position of a given pk in edges
+func (se *SignedEntry) Index(pk cipher.PubKey) byte {
```
evanlinjin: I think it will be a good security measure to also compare `pk` with the other edge, and have a second output (of `error`), just in case of `pk` not equivalent to both edges.

pkg/transport/entry.go
```diff
@@ -70,6 +95,33 @@ type SignedEntry struct {
...
+}
+
+// SetSignature sets Signature for a given PubKey in correct position
+func (se *SignedEntry) SetSignature(pk cipher.PubKey, secKey cipher.SecKey) {
+	idx := se.Index(pk)
+	se.Signatures[idx] = se.Entry.Signature(secKey)
+}
+
+// GetSignature gets Signature for a given PubKey from correct position
+func (se *SignedEntry) GetSignature(pk cipher.PubKey) cipher.Sig {
```
evanlinjin: I think we should avoid the verb `Get` (as from my understanding, Golang convention suggests avoiding it). I think `Signature` will be fine here.

pkg/transport/handshake.go
```diff
@@ -30,21 +28,22 @@ func (handshake settlementHandshake) Do(tm *Manager, tr Transport, timeout time.
...
-			entry.ID = uuid.New()
+		// sEntry := &SignedEntry{Entry: entry, Signatures: [2]cipher.Sig{entry.Signature(tm.config.SecKey)}}
+
+		sEntry := NewSignedEntry(entry, tm.config.PubKey, tm.config.SecKey)
+		if err := validateSignedEntry(sEntry, tr, tm.config.PubKey); err != nil {
+			return nil, fmt.Errorf("NewSignedEntry: %s", err)
 		}
 
-		sEntry := &SignedEntry{Entry: entry, Signatures: [2]cipher.Sig{entry.Signature(tm.config.SecKey)}}
 		if err := json.NewEncoder(tr).Encode(sEntry); err != nil {
```
evanlinjin, line 48: We may need to encode/decode into different structures. That way we can compare between the two, and ensure that only a signature is changed.

pkg/transport/manager.go
```diff
@@ -141,12 +141,29 @@ func (tm *Manager) ReconnectTransports(ctx context.Context) {
...
+// Remote returns the key from the edges that is not equal to Manager.config.PubKey
+// in case when both edges are different - returns empty cipher.PubKey{}
+func (tm *Manager) Remote(edges [2]cipher.PubKey) cipher.PubKey {
+	if tm.config.PubKey == edges[0] {
+		return edges[1]
+	}
+	if tm.config.PubKey == edges[1] {
+		return edges[0]
+	}
+	return cipher.PubKey{}
```
evanlinjin: Instead of returning an empty `cipher.PubKey`, consider returning an `error`. Thoughts @ayuryshev ?

evanlinjin (changes requested):
Beautiful work so far. Keep the changes coming!

pkg/transport/entry.go
ayuryshev: Just `Sign`?
Actually I thought about limiting signing to just 2 operations: 
- on init
- on respond

and do nothing in case when signature is already set.

So maybe: InitSign - ResponseSign ?
Code will became more readable - possible errors will catch the eye immediately

Changing to Sign

pkg/transport/entry.go
ayuryshev: I thought about it.

It makes impossible to write in such way:

```go
sEntry.Signatures[sentry.Index(pk)]
```

I thought about alternatives:

- `(byte,error)` in result
- return an impossible value, e.g. `3` or '255' - must lead to "interesting" errors 
- and even `panic` (that thought was thrown away immediately)

So ..
It's easy to change - do so?

pkg/transport/entry.go
ayuryshev: Agreed

pkg/transport/manager.go
ayuryshev: It's the same as with `SignedEntry.Index()`

OK.

I will change to `(cipher.PubKey, error)` this
And `SignedEntry.Index()` to `(byte, error)`

It will make code a little bulkier but more "golangish".

pkg/transport/entry.go
evanlinjin: I would prefer`SignAsResponder` and `SignAsInitiator`. Either that or a single function `Sign`.

pkg/transport/entry.go
evanlinjin: We don't need to write code like that. It's more important to handle errors and unexpected behaviors properly.

pkg/transport/manager.go
evanlinjin: "golangish" is the way of the Jedi. May the force be with you.

pkg/transport/handshake.go
ayuryshev: OK
I did exactly so in tests.
Here it remained as was before.
Agreed

pkg/node/rpc.go
```diff
@@ -69,17 +69,20 @@ type TransportSummary struct {
...
-	summary := TransportSummary{
-		ID:     tp.ID,
-		Local:  tp.Local(),
-		Remote: tp.Remote(),
-		Type:   tp.Type(),
-	}
-	if includeLogs {
-		summary.Log = tp.LogEntry
+func newTransportSummary(tm *transport.Manager, tp *transport.ManagedTransport, includeLogs bool) *TransportSummary {
+	if remote, Ok := tm.Remote(tp.Edges()); Ok == nil {
```
ivcosla: why not `ok` instead?

pkg/node/rpc_client.go
```diff
@@ -314,7 +314,7 @@ func (mc *mockRPCClient) Transport(tid uuid.UUID) (*TransportSummary, error) {
 // AddTransport implements RPCClient.
 func (mc *mockRPCClient) AddTransport(remote cipher.PubKey, tpType string, public bool, _ time.Duration) (*TransportSummary, error) {
 	summary := &TransportSummary{
-		ID:     uuid.New(),
+		ID:     transport.MakeTransportID(mc.s.PubKey, remote, tpType),
```
ivcosla: I would name it `NewID`, since it's already a function of `transport` package.

ivcosla:
Seems to fail in an infinite loop when starting a node that previously had transports registered with `mainnet` branch.

Currently apps don't work. Loop creation fails with the following logs:

```
[2019-04-04T16:53:33+02:00] WARN [trmanager]: Failed to accept connection: Failed to recover pubkey from signature
[2019-04-04T16:54:34+02:00] INFO [router]: App request with type CreateLoop failed: route setup: rule setup: transport: deadline exceeded
```

pkg/manager/user.go
```diff
@@ -26,7 +26,7 @@ func init() {
 // User represents a user of the manager.
 type User struct {
 	Name   string
-	PwSalt []byte
+	PwSalt []byte /*  */
```
ivcosla: Intended?

pkg/messaging/client.go
```diff
@@ -49,9 +49,11 @@ type Config struct {
 }
 
 // Client sends messages to remote client nodes via relay Server.
+// Implements Transport
 type Client struct {
 	Logger *logging.Logger
 
+	// edges [2]cipher.PubKey
```
ivcosla: Rather remove it than leave it commented, unless you have a reason

pkg/messaging/client.go
```diff
@@ -142,6 +144,7 @@ func (c *Client) Accept(ctx context.Context) (transport.Transport, error) {
 func (c *Client) Dial(ctx context.Context, remote cipher.PubKey) (transport.Transport, error) {
 	entry, err := c.dc.Entry(ctx, remote)
 	if err != nil {
+		fmt.Printf("Dial with remote = %v\n", remote)
```
ivcosla: Debug comment?

pkg/router/route_manager.go
```diff
@@ -74,10 +74,10 @@ func (rm *routeManager) RemoveLoopRule(addr *app.LoopAddr) error {
 }
 
 func (rm *routeManager) Serve(rw io.ReadWriter) error {
-	proto := setup.NewProtocol(rw)
-	fmt.Println("got proto!")
-
+	proto := setup.NewSetupProtocol(rw)
 	t, body, err := proto.ReadPacket()
+
+	fmt.Println("got proto!")
```
ivcosla: same

pkg/router/route_manager.go
```diff
@@ -74,10 +74,10 @@ func (rm *routeManager) RemoveLoopRule(addr *app.LoopAddr) error {
...
 	if err != nil {
 		fmt.Println("err:", err)
```
ivcosla, line 81: here too

pkg/transport/entry_test.go
```diff
@@ -0,0 +1,164 @@
...
+	} else {
+		fmt.Printf("sEntry.Signatures:\n%v\n", sEntry.Signatures)
+	}
+
+	// Output: No signatures set
+	// One signature set
+	// Both signatures set
+}
+
+func errorsPrint(errs ...error) {
```
ivcosla: a minor thing, but guess `printErrors` would be better

pkg/transport/handshake.go
```diff
@@ -30,38 +28,47 @@ func (handshake settlementHandshake) Do(tm *Manager, tr Transport, timeout time.
...
+			ID:       MakeTransportID(tr.Edges()[0], tr.Edges()[1], tr.Type(), public),
+			EdgeKeys: tr.Edges(),
+			Type:     tr.Type(),
+			Public:   public,
 		}
 
-		newEntry := id == uuid.UUID{}
-		if newEntry {
-			entry.ID = uuid.New()
+		// sEntry := &SignedEntry{Entry: entry, Signatures: [2]cipher.Sig{entry.Signature(tm.config.SecKey)}}
```
ivcosla: This should be removed

pkg/transport/manager_test.go
```diff
@@ -133,27 +135,33 @@ func TestTransportManagerReEstablishTransports(t *testing.T) {
...
 	require.NoError(t, err)
-	assert.False(t, dEntry.IsUp)
+	assert.False(t, dEntry2.IsUp)
 
 	m2, err = NewManager(c2, f2)
 	require.NoError(t, err)
-	go m2.Serve(context.TODO()) // nolint
 
-	time.Sleep(time.Second)
+	// errCh2 := make(chan error)
```
ivcosla: remove it

pkg/transport/tcp_transport.go
```diff
@@ -77,19 +77,25 @@ func (f *TCPFactory) Type() string {
 // TCPTransport implements Transport over TCP connection.
 type TCPTransport struct {
 	*net.TCPConn
-	lpk cipher.PubKey
-	rpk cipher.PubKey
+	edges [2]cipher.PubKey
+	// lpk cipher.PubKey
```
ivcosla: same for these two

pkg/transport/tcp_transport_test.go
```diff
@@ -55,8 +55,8 @@ func TestTCPFactory(t *testing.T) {
 	tr, err := f2.Dial(context.TODO(), pk1)
 	require.NoError(t, err)
 	assert.Equal(t, "tcp", tr.Type())
-	assert.Equal(t, pk2, tr.Local())
-	assert.Equal(t, pk1, tr.Remote())
+	// assert.Equal(t, pk2, tr.Local())
```
ivcosla: again

pkg/node/rpc_client.go
ayuryshev: `MakeTransportID` was an @evanlinjin  suggestion.
And in transport we have types:

- DiscoveryClient interface 
- Entry struct 
- EntryWithStatus struct 
- Factory interface 
- LogEntry struct 
- LogStore interface 
- ManagedTransport struct 
- Manager struct 
- ManagerConfig struct 
- MockFactory struct 
- MockTransport struct 
- PubKeyTable interface 
- SignedEntry struct 
- Status struct 
- TCPFactory struct 
- TCPTransport struct 
- Transport interface 

So I think `MakeTransportID` makes sense

pkg/manager/user.go
ayuryshev: No. Clearing

pkg/messaging/client.go
ayuryshev: Clearing

pkg/transport/handshake.go
ayuryshev: Yes

pkg/transport/manager_test.go
ayuryshev: Yes

pkg/transport/tcp_transport_test.go
ayuryshev: yes again

pkg/messaging/channel.go
```diff
@@ -29,6 +30,10 @@ type channel struct {
 	noise *noise.Noise
 }
 
+func (c *channel) Edges() [2]cipher.PubKey {
```
evanlinjin, line 34: We need a comment here.

pkg/node/rpc.go
ayuryshev: It's already resolved

pkg/messaging/client.go
ayuryshev: Yes

pkg/router/route_manager.go
ayuryshev: Yes

pkg/transport/entry.go
```diff
@@ -26,6 +27,31 @@ type Entry struct {
...
+func (e *Entry) Edges() [2]cipher.PubKey {
+	// this sort *must* be needless
+	// but to remove it:
+	// - all tests must be passed
+	// - written Benchmarks
+	return SortPubKeys(e.EdgeKeys[0], e.EdgeKeys[1])
+}
+
+// SetEdges sets edges of Entry
+func (e *Entry) SetEdges(edges [2]cipher.PubKey) {
```
evanlinjin, line 45: Instead of `edges [2]cipher.PubKey` as input, I think `edge1, edge2 cipher.PubKey` would be clearer.

pkg/transport/entry_test.go
ayuryshev: It's almost stylistical.
What feels more natural: SVO or SOV?
So ..
It's used only one time - so I will make it anonimous:

```go
	doneCh := make(chan bool)
	go func(errs ...error) {
		for _, err := range errs {
			if err != nil {
				fmt.Println(err.Error())
			}
		}
		doneCh <- true
	}(errIdxA, errIdxB, errSigA, errSigB)

	<-doneCh
```

pkg/transport/manager_test.go
ayuryshev: Done

pkg/transport/entry.go
```diff
@@ -70,6 +97,44 @@ type SignedEntry struct {
 	Registered int64         `json:"registered,omitempty"`
 }
 
+// Index returns position of a given pk in edges
+func (se *SignedEntry) Index(pk cipher.PubKey) (byte, error) {
```
evanlinjin: Instead of returning an error, I think returning a boolean will make more sense. Take a note of it in the comments: "returns false if public key is not found".

pkg/transport/entry.go
```diff
@@ -70,6 +97,44 @@ type SignedEntry struct {
...
+		return 1, nil
+	}
+	if pk == se.Entry.Edges()[0] {
+		return 0, nil
+	}
+	return 0xff, errors.New("invalid pubkey")
+}
+
+// Sign sets Signature for a given PubKey in correct position
+func (se *SignedEntry) Sign(pk cipher.PubKey, secKey cipher.SecKey) error {
```
evanlinjin: return a boolean. "Returns false if signature is not set".

pkg/transport/entry.go
```diff
@@ -70,6 +97,44 @@ type SignedEntry struct {
...
+	idx, err := se.Index(pk)
+	if err == nil {
+		se.Signatures[idx] = se.Entry.Signature(secKey)
+	}
+	return err
+}
+
+// Signature gets Signature for a given PubKey from correct position
+func (se *SignedEntry) Signature(pk cipher.PubKey) (cipher.Sig, error) {
```
evanlinjin: Return a boolean. "Returns false if no signature is found of given public key."

pkg/transport/handshake.go
```diff
@@ -30,38 +28,45 @@ func (handshake settlementHandshake) Do(tm *Manager, tr Transport, timeout time.
...
+			return nil, fmt.Errorf("settlementInitiatorHandshake NewSignedEntry: %s\n sEntry: %v", err, sEntry)
 		}
 
-		sEntry := &SignedEntry{Entry: entry, Signatures: [2]cipher.Sig{entry.Signature(tm.config.SecKey)}}
 		if err := json.NewEncoder(tr).Encode(sEntry); err != nil {
 			return nil, fmt.Errorf("write: %s", err)
 		}
 
-		if err := json.NewDecoder(tr).Decode(sEntry); err != nil {
+		rcvdSEntry := &SignedEntry{}
```
evanlinjin: maybe `respEntry` will be better?

evanlinjin: We also need to compare the sent Entry with the Received Entry and ensure no fields are modified.

pkg/transport/manager.go
```diff
@@ -141,12 +143,30 @@ func (tm *Manager) ReconnectTransports(ctx context.Context) {
...
 }
 
+// Local returns Manager.config.PubKey
+func (tm *Manager) Local() cipher.PubKey {
+	return tm.config.PubKey
+}
+
+// Remote returns the key from the edges that is not equal to Manager.config.PubKey
+// in case when both edges are different - returns empty cipher.PubKey{}
+func (tm *Manager) Remote(edges [2]cipher.PubKey) (cipher.PubKey, error) {
```
evanlinjin: I would return `bool` instead of `error`.

evanlinjin (changes requested):
Looks really good. Please review my queries.

pkg/transport/entry.go
evanlinjin: Ignore this.

pkg/transport/entry_test.go
evanlinjin: I think if we have a separate function, it would make the code easier to read.

ayuryshev: This function is used just one time - that's the reason why I changed it anonimous function.

In case you propose to change it to named function - please propose name for it.
Be it: 
- printErrors
- errPrint
- eprint

I'm OK with each choice

evanlinjin: `printErrors`. We can have it as a var within the function. `printErrors := func() ....`

pkg/messaging/channel.go
ayuryshev: Yes
```
func (c *channel) Edges() [2]cipher.PubKey {
	// Edges returns the public keys of the Transport's edge nodes (should only have 2 edges and the least-significant edge should come first).
	return transport.SortPubKeys(c.link.Local(), c.remotePK)
}
```
It will be clearer this way

evanlinjin: It's golang standard to have a comment above each exposed function. I prefer we stick to the standard.

pkg/transport/entry.go
ayuryshev: We can follow the lead by standard libs and return `-1` instead of `error`.
E.g. <https://golang.org/pkg/strings/#Index>

Is it better?  - IDK

I can change it to do in such way

pkg/transport/entry.go
ayuryshev: OK

pkg/transport/entry.go
ayuryshev: OK

pkg/transport/entry.go
evanlinjin: > We can follow the lead by standard libs and return `-1` instead of `error`.
> E.g. https://golang.org/pkg/strings/#Index
> 
> Is it better? - IDK
> 
> I can change it to do in such way

Good idea. Yes.

evanlinjin: instead of `byte`, we will need `int8`.

pkg/transport/manager.go
ayuryshev: Again: maybe follow the lead by standard libs and return `-1`?

evanlinjin: I think we should stick to `bool`. As we are returning a `cipher.PubKey`.

pkg/transport/entry.go
ayuryshev: OK.
I'm changing it to `-1` logic - it will make code "lighter" in most cases and for rare cases we will need checks for `-1`
As with standard libs

pkg/transport/handshake.go
ayuryshev: agreed for respEntry

We already have this check in `validateSignedEntry`:

```go
	if entry.Edges() != tr.Edges() {
		return errors.New("invalid entry edges")
	}
```

pkg/transport/manager.go
ayuryshev: Agreed
Such code looks more natural this way:
```go
	r.node.tm.WalkTransports(func(tp *transport.ManagedTransport) bool {
		if remote, ok := r.node.tm.Remote(tp.Edges()); ok {
			if typeIncluded(tp.Type()) && pkIncluded(r.node.tm.Local(), remote) {
				*out = append(*out, newTransportSummary(r.node.tm, tp, in.ShowLogs))
			}
			return true
		}
		return false
	})
```

pkg/messaging/channel.go
evanlinjin: https://blog.golang.org/godoc-documenting-go-code

Please place the comment above the function. Thanks.

pkg/router/route_manager.go
evanlinjin: Logging should be done via `github.com/skycoin/skycoin/src/util/logging`, not `fmt`.

pkg/transport/tcp_transport.go
evanlinjin: Please fix this.

evanlinjin (changes requested):

pkg/messaging/channel.go
ayuryshev: Yes

pkg/transport/tcp_transport.go
ayuryshev: Done

ayuryshev: Done

pkg/messaging/channel.go
ayuryshev: yes

evanlinjin (approved):
I have no way to test this code in an easy manner until the local testing environment is done. I will approve this now and deal with possible complications later. 🙂

Will merge after @ivcosla approves.

PR #257, opened 2019-03-29
Refactored services to have reusable config file logic.

### LINKED ISSUE #255: Create shared library for discovering config files.
Currently we have an implementation in `/internal/pathutil`. However, this comes with some level of restriction, and not all services currently make use of it.

## Tasks
- [x] Make `pathutil.Find()` function more customisable.
- [x] Make `manager-node` use `pathutil`.

evanlinjin:
closes #255


ivcosla (approved):
I like it quite a lot!

PR #265, opened 2019-04-01
sort transports listed by skywire-cli list-transports command

### LINKED ISSUE #250: Sort skywire-cli list-transports
Currently `skywire-cli list-transports` return the transport list in the map key order, which is random, it would be better for usability purpose to return it sorted.

I suggest sort it by transport id.

mungujn:
fixes #250 
I have added a method that sorts transports by transport-id before printing them out


evanlinjin (approved):
So beautiful I want to cry.

PR #268, opened 2019-04-04
fixed ack issue

### LINKED ISSUE #225: ACK failure upon many send operations
I found this error by using chat app and writing many messages, eventually there is a CRC error within ACK: https://github.com/skycoin/skywire/blob/mainnet/internal/ioutil/ack.go#L194

Errors shown on both nodes:
Sender
Captura de pantalla 2019-03-15 a las 10 56 31

Receiver
Captura de pantalla 2019-03-15 a las 10 56 47

After such error it is impossible to keep using chat app.

ivcosla:
Closes #225.

There is a modification to data buffer in line https://github.com/skycoin/skywire/blob/b85aee6f251fa7aa5d8a6641f64b8c4bb11f2ce6/internal/ioutil/ack.go#L146

The previous line was before launched in a goroutine, which often executed after such modification, leading to a comparison between wrong ack's, and thus the reported failure.

A similar behavior can be achieved by adding a wait group and keeping the `confirm` function executing in a goroutine, but current implementation should not have a noticeable performance impact.

### DIFF
--- a/internal/ioutil/ack.go
+++ b/internal/ioutil/ack.go
@@ -192,6 +192,7 @@ func (arw *AckReadWriter) confirm(seq byte, hash []byte) {
 
 	if ack.hash != rcvHash {
 		ack.errChan <- errors.New("invalid CRC")
+		return
 	}
 
 	ack.errChan <- nil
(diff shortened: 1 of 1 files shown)


evanlinjin (approved):
Beautiful.

Issue #282, opened 2019-04-12
Fix skywire-node CLI


ivcosla:
Last CLI changes added `vendor` sub-command and now when the path to the json config is passed to the CLI it is taken as a command instead, thus making it fail.

Possible quick fixes:

1. Add config path as a flag, using a default value.
2. Add a new command `start` and take path as an argument.

PR #283, opened 2019-04-12
Fix skywire-cli. Update Makefile

ayuryshev:
Changes:

1. skywire-node: removed command 'version' instead of this used builtin
cobra ability '--version'
2. Makefile: updated for changes in skywire-cli

Closes: #282


ivcosla (approved):
Good job!

PR #299, opened 2019-04-19
Feature/add setup node

ivcosla:
Added setup node to skywire


evanlinjin (approved):
So beautiful I want to cry.

pkg/setup/node_test.go
```diff
@@ -173,11 +178,11 @@ func TestCloseLoop(t *testing.T) {
 	require.NoError(t, err)
 
 	n3 := newMockNode(m3)
-	go n3.Serve(context.TODO()) // nolint: errcheck
+	go n3.serve() // nolint: errcheck
 
 	time.Sleep(100 * time.Millisecond)
 
-	sn := newMockNode(mS)
+	sn := &Node{logging.MustGetLogger("routesetup"), mS, nil, 0, metrics.NewDummy()}
```
ayuryshev, line 183: I see here change `newMockNode` to `&Node` but only here.
Do `mockNode` needed at all?

ayuryshev (approved):

pkg/setup/node_test.go
ivcosla: Yes, it's convenient and I want to keep it, since it allows to customize logic inside skywire node.

PR #300, opened 2019-04-19
reverted json encoding

ivcosla:


evanlinjin (approved):
Although this fixes the problem, this makes our JSON field names inconsistent. We are split between; camelcase, dashes and also underscore...

Also note that I think we have decided to do PRs from forks, instead of creating branches within `skycoin/skywire`.

Issue #308, opened 2019-04-25
Write tests for the new `router.ProcManager` structure.


evanlinjin:
Branch: [`mainnet-dev`](https://github.com/skycoin/skywire/tree/mainnet-dev)

## Prerequisites

1. Move `genMockApp()` from test file of `/pkg/app` module to `/pkg/app/apptest`: https://github.com/skycoin/skywire/blob/64860b09f102a34d78b6eae80efcaa5d7c7cba66/pkg/app/executor_test.go#L29-L49

2. Create an empty implementation of `router.Router`.

## Tests

- [ ] `RunProc()` should:
    - Fail if input `c.BinLoc` is invalid.
    - Succeed succeed if all other inputs are valid.
    - Not assign a port if input `port=0`.
    - After calling `RunProc()`, the returned `*AppProc` should:
        - Have `ProcID()` return `1` on first process, `2` in second, etc.
        - Have `Stopped()` return `false`.
        - After calling `Stop()`, should return no error and `Stopped()` should then return `true`.

- [ ] `Proc()` should return the right `*AppProc` of given `ProcID`, and should return false if:
    - `RunProc` never returned such a `ProcID`.
    - `AppProc` is stopped.

- [ ] `AllocPort()` should always return free ports.

- [ ] `ProcOfPort()` should return `proc` assigned to given port only if `proc` is still running (not `.Stopped()`).

- [ ] `RangeProcIDs()` and `RangePorts()` should cover all that is expected.

- [ ] `Close()` should succeed.

evanlinjin:
@ivcosla I've started work on this. I finished implementing two tests and wrote barebones and comments for subsequent tests to be written.

**Please refer to PR: #317**

PR #317, opened 2019-04-29
Initial tests for router.ProcManager.

evanlinjin:
> #308
- Wrote tests MustFailWithInvalidBinLoc and MustSucceedWithValidValues
- Added TODOs for additional tests to be written.

### DIFF
--- a/pkg/router/procmanager.go
+++ b/pkg/router/procmanager.go
@@ -91,58 +91,58 @@ func NewAppProc(pm ProcManager, r Router, pid ProcID, m *app.Meta, c *app.ExecCo
 }
 
 // ProcID returns the process ID.
-func (ar *AppProc) ProcID() ProcID {
-	return ar.pid
+func (ap *AppProc) ProcID() ProcID {
+	return ap.pid
 }
 
 // Stopped returns true if process is stopped.
-func (ar *AppProc) Stopped() bool {
-	return *(*bool)(atomic.LoadPointer(&ar.stopped))
+func (ap *AppProc) Stopped() bool {
+	return *(*bool)(atomic.LoadPointer(&ap.stopped))
 }
 
 // Stop sends SIGTERM to the hosted App, closes all loops and ends handling of packets.
 // An error is returned if app does not end cleanly, or if app already stopped.
-func (ar *AppProc) Stop() error {
-	ar.mx.Lock()
-	defer ar.mx.Unlock()
+func (ap *AppProc) Stop() error {
+	ap.mx.Lock()
+	defer ap.mx.Unlock()
 
 	// load 'true' to 'stopped'.
 	t := true
-	atomic.StorePointer(&ar.stopped, unsafe.Pointer(&t)) //nolint:gosec
+	atomic.StorePointer(&ap.stopped, unsafe.Pointer(&t)) //nolint:gosec
 
-	ar.log.Info("stopping...")
+	ap.log.Info("stopping...")
 
-	for lm := range ar.lps {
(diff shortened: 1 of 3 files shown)


ayuryshev (commented):
>> ./setup_node binary is not `.gitignore`-d
I was wrong here. I see it's in deleted

evanlinjin:
> ./setup_node binary is not `.gitignore`-d
> I was wrong here. I see it's in deleted

No worries man, it happens ☺️

pkg/router/procmanager.go
```diff
@@ -209,10 +203,41 @@ func (ar *AppProc) ConsumePacket(lm app.LoopMeta, ciphertext []byte) error {
 		return fmt.Errorf("%s: %s", ErrDecryptionFailed.Error(), err.Error())
 	}
 	df := &app.DataFrame{Meta: lm, Data: plaintext}
-	_, err = ar.e.Call(appnet.FrameData, df.Encode())
+	_, err = ap.e.Call(appnet.FrameData, df.Encode())
 	return err
 }
 
+func (ap *AppProc) makeDataHandlerMap() appnet.HandlerMap {
```
ayuryshev, line 210: I have question about this structure:
- It's a  map of closure funcs
- than it used as `handle, ok := handlerMap[t]`

So in essence it's a dynamic dispatch.
Why not pack them into structure and use static dispatch?

ayuryshev (commented):
Great! So ProcManager test are near to finish!

pkg/router/procmanager.go
evanlinjin: How is it a dynamic dispatch?

evanlinjin: There are no `interface`s here.

evanlinjin:
@ayuryshev not quite finished. There are lots of `TODO`s.

pkg/router/procmanager.go
evanlinjin: @ayuryshev we need to handle the incoming frames based on the frame type. Why are changes needed and how does it help?

ayuryshev: We have only 6 appnet.FrameType overall
And 3 in `makeDataHandlersMap`

We can use regular struct (e.g. `dataHandlers`) instead of map 
And make those handlers are regular methods of this structure

This way calls will be dispatched statically: `dh.requestLoop`, `dh.closeLoop`

And with  we have:
- lookup in map for function
- call function

evanlinjin: Okay, this is most definitely a separate issue. The issue you are pointing out regards performance; is this correct?

ayuryshev: Althought it affects performance,  question is not about performance - it's about clarity.

I see that all three closure funcs could be methods of `AppProc`
Why not make them in such way?

At least it will make them more discoverable than:

```golang
		default:
			handle, ok := handlerMap[t]
                        .....                   
                        // What is called here?
                         go func(handle HandlerFunc, payload []byte) {
```

evanlinjin: @ayuryshev I agree that it is not the most clear of code. But it is readable. I prefer if we leave it as is for now, as we will replace `x.Protocol` with `rpc` in the future. What do you think?

ayuryshev: Definitely, we can leave it as it is right now
And I propose to make it "more static" in future

evanlinjin: @ayuryshev thanks for pointing that out. Will keep that in mind.

PR #319, opened 2019-04-29
Feature/tests for new router 308

ayuryshev:
**Request for comments**

1. Created tests for pkg/router. 

Problems exists:
- some of them are problem of test environment
- and some are problems discovered by tests

Discovered problems:

- `Example_setupHandlers_confirmLoop` in **most** (but not all) cases
finishes with "chacha20poly1305: message authentication failed"
- `logging.LevelFromString` + `logging.SetLevel`  causes data races
- ProcManager.RunProc - sometimes panics
- periodical problems arises **after** the test - Close/Stop/TearDown
logic is not very clear

2. Created mini-framework for creating environment, running tests in
them, stopping/tearing down them

Typical test looks like:

```go
	env := &TEnv{}
	_, err := env.RunAsExample(true,
		GenerateKeys(),
		AddTransportManagers(),
		AddProcManagerAndRouter(),
		AddSetupHandlersEnv(),
		// Preparation for confirmLoop
		AddRules(),
		AddLoopData(),
		AddAppAndRunProc("/bin", "sh"), // This step is unstable
		//confirmLoop entrails
		CheckRulesAndPorts(),
		ConfirmLoopAndFinish(),
	)
```

Idea for such test composition came from Rob Pike: ["Self-referential functions and the design of options"](https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html)

Look at `pkg/router/tenv_test.go` and `pkg/router/cfgsteps_test.go`

Proposal: 
- move it out of router packager into separate package
- use this package for tests requiring complex configurations

3. Tests for `router.ProcManager` & `router.RoutingTableManager` 

Although it was not a goal for this PR this structures are partially tested for obvious reasons - `router` depends on them. 

Sad part - most of the coverage are from no_ci-tests.


pkg/router/router-handlers.go
```diff
@@ -0,0 +1,48 @@
...
+	setupPKs := make(map[cipher.PubKey]struct{})
+	for _, pk := range r.conf.SetupNodes {
+		setupPKs[pk] = struct{}{}
+	}
+
+	return routerHandlers{r, pm, setupPKs}
+}
+
+// determines if the given transport is established with a setup node.
+func (rh routerHandlers) isSetup() func(tp transport.Transport) bool {
```
evanlinjin: I think this should be a member of `router` and there is no need to further return a function.
```golang
func (r router) isSetup(tp transport.Transport) bool { /* ... */ }
```

pkg/router/router-handlers.go
```diff
@@ -0,0 +1,48 @@
+package router
+
+import (
+	"io"
+
+	"github.com/skycoin/skywire/pkg/cipher"
+	"github.com/skycoin/skywire/pkg/transport"
+)
+
+type routerHandlers struct {
```
evanlinjin: Maybe `transportHandlers` will be better?

pkg/router/router-handlers.go
```diff
@@ -0,0 +1,48 @@
...
+			return false
+		}
+		_, ok = rh.setupPKs[pk]
+		return ok
+	}
+}
+
+// serves a given transport with the 'handler' running in a loop.
+// the loop exits on error.
+func (rh routerHandlers) serve() func(tp transport.Transport, handle func(ProcManager, io.ReadWriter) error) {
```
evanlinjin: Don't return a function here please, it is not needed.

pkg/router/router.go
```diff
@@ -65,34 +64,14 @@ type router struct {
 	rfc  routeFinder.Client
 }
 
+func (r *router) String() string {
```
evanlinjin: Just curious what the purpose of this is?

pkg/router/router-handlers.go
evanlinjin: Ideally I think it should still be a variable function declared within `router.Serve()`...

pkg/router/router-handlers.go
evanlinjin: However, I don't think I see how this structure helps?

pkg/router/setup-handlers.go
```diff
@@ -0,0 +1,160 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/skycoin/skywire/pkg/routing"
+	"github.com/skycoin/skywire/pkg/setup"
+)
+
+type setupHandlers struct {
```
evanlinjin: Personally I don't think the nested functions in `handleSetup` cause any problems with testability. All that is needed to test a function is to send data into `rw io.ReadWriter` with a certain data type.

evanlinjin: If testing is difficult, I prefer a simpler solution of just moving these nested functions out. I'm curious to know how you think an introduction of `setupHandlers` helps?

evanlinjin (commented):
@ayuryshev you may need to justify many of your decisions here.

pkg/router/router-handlers.go
ayuryshev: My decisions was based on principle: 

"While it's not tested - don't change behaviour. Let it remain as is"

So: 
- the closures remains closures
- no new methods for existing types from start - it could be refactored later 
- slight change: each closure created on fly only when needed instead of creating all of them upfront

This is same for router-handlers and setup-handlers

pkg/router/router-handlers.go
ayuryshev: `transportHandlers` does not sounds right here:

- if innards of `handleSetup` were extracted into setupHandlers
- then logically `transportHandlers` must contain innards of `handleTransport` 

and that's not so

ayuryshev: About overall benefits of this structure - it's not clear for me too right now
It's WIP

pkg/router/router-handlers.go
ayuryshev: As I explained above - this come from principle "Don't change behaviour on start"
So closure remains closure. 
It's only exposed

pkg/router/router.go
ayuryshev: It was a failed attempt to create good-looking string representation of `router`
Removing

pkg/router/setup-handlers.go
ayuryshev: Please look at last commit

It's certainly adds testability IMO

And  I cannot prove this hypothesis( because we don't have Benchmarks now): 
- It must be slightly faster (closure created only when needed)

pkg/router/router-handlers.go
ayuryshev: Changed to regular methods

pkg/router/setup-handlers.go
ayuryshev: Changed to regular methods

pkg/router/router-handlers.go
evanlinjin: > Ideally I think it should still be a variable function declared within `router.Serve()`...

Can we move this back as a closure in `router.Serve`?

pkg/router/router-handlers.go
evanlinjin: > About overall benefits of this structure - it's not clear for me too right now
> It's WIP

@ayuryshev what are you trying to achieve with this structure?

pkg/router/router-handlers.go
ayuryshev: 5 days later.
Please look again in code. 

The structure itself is: 
- purely supplementary
- it's possible to make those handlers as methods of `router` 
- and `serve` receiving ProcManager as argument 

Isolating them as methods of separate structure:
- makes interface of router "lighter"
- it's the same as principle as `setupHandlers` for `router` and "handlerMap"s for `ProcManager`
- possible performance cost must be negligible

Let's discuss
I'm ready to refactor them into router-methods

pkg/router/router-handlers.go
ayuryshev: I explained already that this structure was created purely mechanically: as extract of Serve-function namespace that was needed to move closures outside of `Serve`.

As for now: I'm not "pro-" or "contra-" for this.

evanlinjin (commented):
Instead of commenting on the code, I would like to provide some general points.
1. This should have really been split into multiple PRs relating to multiple issues.
2. The refactoring of the production code, although well-intentioned, makes little sense to me despite the explanations behind it.
3. The test-environment structure is a great idea. However, I'm not a fan of the use of functional arguments.
  - We do not need this to be able to set up the test environment and this diverges from the style of tests is `skycoin/skywire` as well as `skycoin/skycoin`.
  - I've noticed that you started using functional arguments as sub-tests as well. Sub-tests should we run with `t.Run()`.
  - Calls to `assert` and `require` libraries should be kept within the scope of the test.
4. The use of examples as replacement for comments and tests is unacceptable.
5. In the future, we require each test to have comments describing the intention of the test and what we are testing.

As a plan of action, the following steps will be taken:
1. This PR is to be "scrapped", although parts of this codebase can be reused.
2. We need to have the changes in `/pkg/setup` in a separate PR.
3. As I agree that we need to improve the test-ability of the `router` library, I will refactor the code to suit my tastes (and provide these changes in a separate PR). From there, we can point out problems with my suggested approach.
4. We need to create a document to plan and describe tests that are to be written to ensure complete testing of `/pkg/router` package.
5. From there, we will formulate tasks for different people.

evanlinjin:
Refer to comment: https://github.com/skycoin/skywire/pull/319#pullrequestreview-234842217

Issue #321, opened 2019-05-01
Re-implement app-related commands in skywire-cli and manager-node.


evanlinjin:
> Branch: `mainnet-dev`

## Background
After the separation of the proc and app concepts (Where a proc is an instance of a running app), the interfaces; `node.RPC` and `node.RPCClient` (as well as their implementations) need an update. And consequently, also the endpoints of `manager.Node` and `skywire-cli`.

## Tasks
- [x] Replace the commented out methods of `node.RPC`/`node.RPCClient` with the following methods:
  - `Apps() []app.Meta` obtains all available apps.
  - `StartProc(appName string, args []string, port uint16) (router.ProcID, error)` runs a proc.
  - `StopProc(pid router.ProcID) error` stops a proc.
  - `ListProcs() []router.ProcInfo` lists all running procs. Note that `ProcInfo` is not implemented; it should be a combination of `router.ProcID` ,`app.Meta` and `app.ExecConfig`.

- [x] Write sub-commands for the aforementioned calls in `skywire-cli`.
- [ ] Write endpoints for the aforementioned calls in `manager.Node`.

PR #326, opened 2019-05-01
added rpc methods for procs

ivcosla:
Added methods to RPC and ProcManager. Related to #321


pkg/app/executor.go
```diff
@@ -94,6 +94,22 @@ func (ec *ExecConfig) AppName() string {
 	return filepath.Base(ec.BinLoc)
 }
 
+// ExecutorFactory spawns a certain type of Executors
+type ExecutorFactory interface {
```
ivcosla, line 98: Previously, even if `Executor` was an interface there was no way to plug a different implementation of it into `Node`, since `NewExecutor` was always called inside to spawn new ones, thus I deemed necessary to create a factory that Nodes uses to spawn them, allowing other implementations of the factory, which allows to spawn different implementations of `Executor`.

pkg/app/executor.go
```diff
@@ -105,8 +121,11 @@ type Executor interface {
 	Meta() *Meta
 	Call(t appnet.FrameType, reqData []byte) ([]byte, error)
 	CallUI(t appnet.FrameType, reqData []byte) ([]byte, error)
+	SetLogger(logger *logging.Logger)
```
ivcosla, line 124: Before we were creating new executors inside `procManager.RunProc` taking `*app.Meta` and `*app.ExecConfig` as arguments, and a logger was created per app for that executor (also passed as an argument).

Instead I wanted to pass an Executor as argument to RunProc, but needed a way to set such logger, as well as retrieving it, since such logger cannot be created before RunProc is called.

cmd/skywire-cli/commands/node/app.go
```diff
@@ -1,59 +1,99 @@
...
+		for _, meta := range appMetas {
+			_, err = fmt.Fprintf(w, "%s\t%s\t%s\n", meta.AppName, meta.AppVersion, meta.ProtocolVersion)
+			internal.Catch(err)
+		}
+		internal.Catch(w.Flush())
+	},
+}
+
+var startAppCmd = &cobra.Command{
+	Use:   "start-app <name> <port>",
```
evanlinjin: `port` should be a flag and optional. No `--port` specified == use `0` as input for `StartProc` function.

pkg/node/config.go
```diff
@@ -147,11 +147,11 @@ type ManagerConfig struct {
 // AutoStartConfig defines app startup parameters.
 type AutoStartConfig struct {
 	App  string   `json:"app"`
-	Port uint16   `json:"port"`
-	Args []string `json:"args"`
+	Port uint16   `json:"Port"`
```
evanlinjin: Why did we change the tags here?

evanlinjin (changes requested):
I'll complete this review at a later stage.

cmd/skywire-cli/commands/node/app.go
ivcosla: Ok, my mistake, thought we needed different ports every time, that's why I took it as argument.

pkg/node/config.go
ivcosla: That was unintentional, some side effects of renaming.

PR #342, opened 2019-05-09
[M2] [WIP] Feature/implementation of netutil RPCDuplex 298

atang152:
**Summary**
The first structure of the netutil to be implemented is netutil.RPCDuplex. This structure implements an RPC Duplex connection via a single net.Conn implementation where both end of the connection has a client and a server. These two branch connections (server/client) in the RPCDuplex are first prefixed with a header of 3 bytes. The first byte of the header tells the RPCDuplex which branch connection (serverConn or clientConn) to direct to. The second and third byte is the encoded size of the packet. The RPCDuplex will then forward the packet based on it's prefix and size from the original connection to the appropriate prefixed connection.


pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"encoding/binary"
+	"errors"
+	"io"
+	"log"
+	"net"
+)
+
+// PrefixedConn will inherit the net.Conn interface.
+type PrefixedConn struct {
```
evanlinjin: Let's call this `branchConn` (yes, unexported).

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,144 @@
...
+	net.Conn
+	prefix   byte
+	name     string
```
evanlinjin: Please get rid of this, it's not needed.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,144 @@
...
+	readChan chan []byte
```
evanlinjin: Rename to `readCh`.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// ReadHeader reads the first three bytes of the data and returns the prefix and size of the packet
+func (d *RPCDuplex) ReadHeader() (byte, uint16) {
+
+	var bs = make([]byte, 2)
+	var size uint16
+
+	// Read the 1st byte from prefix
+	_, err := d.conn.Read(bs[:1])
```
evanlinjin: I prefer if we read 3 bytes in one go. The less reads the better.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,144 @@
...
+	_, err := d.conn.Read(data)
+	if err != nil {
+		if err != io.EOF {
+			return err
+		}
+		return nil
+	}
+
+	// Push data from original conn into prefixedConn's chan []byte
+	go func() {
```
evanlinjin: Instead of having this logic in a go-routine, we can initiate the associated `readCh`'s with a size (as a constant on top of this file).

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,417 @@
...
+	// 	branchConn:   "clientConn",
+	// 	expectedMsg:  "",
+	// 	expectedSize: uint16(0),
+	// 	expectedConn: "serverConn",
+	// },
+}
+
+func TestNewRPCDuplex(t *testing.T) {
+	for _, tt := range tables {
+		t.Run(tt.description, func(t *testing.T) {
```
evanlinjin, line 270: We need to test repetitive calls to Read/Write of a single pair of `RPCDuplex`. Not multiple pairs. Going along the analogy I've told you before, we are testing multiple cars for one revolution of the wheel. We need to test a single car for multiple revolutions of the wheel.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,417 @@
...
+	bDuplex := NewRPCDuplex(connB, false)
+
+	var wg sync.WaitGroup
+	wg.Add(expectedMsgCount)
+	go func() {
+		for i := 0; i < expectedMsgCount; i++ {
+			msg := fmt.Sprintf("foo%d", i)
+			n, err := aDuplex.clientConn.Write([]byte(msg))
+			// log.Println(msg, n)
+			assert.Nil(err)
```
evanlinjin: It is a bad idea to use `assert` or `require` in separate go routines. Please use a channel.

evanlinjin: If we can use a channel (with size of `expectedMsgCount`) we can also get rid of `wg`.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,417 @@
...
+	connA, connB := net.Pipe()
+	defer connA.Close()
+	defer connB.Close()
+
+	aDuplex := NewRPCDuplex(connA, true)
+	bDuplex := NewRPCDuplex(connB, false)
+
+	msg := []byte("foo")
+	go func() {
+		_, err := aDuplex.clientConn.Write(msg)
```
evanlinjin: Use `errCh := make(chan error, 1)`

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,417 @@
...
+		if err != nil {
+			log.Fatalln("Error writing from conn", err)
+		}
+	}()
+
+	err := bDuplex.Forward()
+
+	bs := <-bDuplex.serverConn.readChan
```
evanlinjin: There is no need for this variable if it is to be only used once.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,417 @@
...
+
+	require.NoError(t, err)
+	assert.Equal(t, []byte("foo"), bs)
+}
+
+// TestPrefixedConn_Read reads data pushed in by
+// the Original connection
+func TestPrefixedConn_Read(t *testing.T) {
+
+	t.Run("successful prefixedConn read", func(t *testing.T) {
```
evanlinjin: Please use `snake_case` for all test names. 🥇

pkg/util/netutil/duplex_test.go
evanlinjin: And use `b` instead of `bs`. `b` is standard in go. `bs` ... you know what that stands for.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,417 @@
...
+		ch := make(chan []byte)
+		pc := &PrefixedConn{prefix: 0, readChan: ch}
+
+		msg := []byte("foo")
+
+		go func() {
+			pc.readChan <- msg
+		}()
+
+		bs := make([]byte, 3)
```
evanlinjin: `b`

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,417 @@
...
+// 2 byte encoded length of the packet.
+func TestPrefixedConn_Write(t *testing.T) {
+
+	connA, connB := net.Pipe()
+	var ch chan []byte
+	defer connB.Close()
+
+	pc := &PrefixedConn{Conn: connA, readChan: ch}
+
+	go func() {
```
evanlinjin: Use channels instead of testing variables in go routine.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,417 @@
...
+		connA, connB := net.Pipe()
+
+		aDuplex := NewRPCDuplex(connA, true)
+		bDuplex := NewRPCDuplex(connB, false)
+
+		msg := []byte("foo")
+
+		go func() {
+			_, err := aDuplex.clientConn.Write(msg)
+			if err != nil {
```
evanlinjin: Channels.

evanlinjin (changes requested):
Good work so far!
After reviewing these changes, please actually put RPC into the mix.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"encoding/binary"
+	"errors"
+	"io"
+	"log"
+	"net"
+)
+
+const readChSize int = 1
```
evanlinjin: Do we need this?

evanlinjin: @atang152 We need this. Please rename this to `bufferSize` and set it to a larger value (ie 8).

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,160 @@
...
+	// branchConn implements net.Conn and assigned it to d.clientConn and d.serverConn
+	if initiator {
+		d.clientConn = &branchConn{Conn: conn, prefix: 0, readCh: clientCh}
+		d.serverConn = &branchConn{Conn: conn, prefix: 1, readCh: serverCh}
+	} else {
+		d.clientConn = &branchConn{Conn: conn, prefix: 1, readCh: clientCh}
+		d.serverConn = &branchConn{Conn: conn, prefix: 0, readCh: serverCh}
+	}
+	d.conn = conn
+	d.rpcS = srv
```
evanlinjin, line 83: We might as well create the `rpc.Client` at this point with [`rpc.NewClient`](https://golang.org/pkg/net/rpc/#NewClient) (having `d.clientConn` as input ofcourse).

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Client returns the internal RPC Client.
+func (d *RPCDuplex) Client() *rpc.Client { return d.rpcC }
+
+// ReadHeader reads the first three bytes of the data and returns the prefix and size of the packet
+func (d *RPCDuplex) ReadHeader() (byte, uint16) {
+
+	var b = make([]byte, 3)
+	var size uint16
```
evanlinjin: We don't need to declare `size` here. It is fine having it on line 65.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,160 @@
...
+func (d *RPCDuplex) Forward() error {
+
+	// Reads 1st byte of prefixed connection to determine which conn to forward to
+	prefix, size := d.ReadHeader()
+
+	// Reads packet with size 'size' from Original Conn into data
+	data := make([]byte, size)
+	_, err := d.conn.Read(data)
+	if err != nil {
+		if err != io.EOF {
```
evanlinjin: Please change to;
```golang
if err == io.EOF { return nil }
```

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,160 @@
...
+			return err
+		}
+		return nil
+	}
+
+	// Push data from original conn into branchConn's chan []byte
+	if prefix == d.serverConn.prefix {
+
+		// Serve conn and init new branch client
+		go d.rpcS.ServeConn(d.clientConn)
```
evanlinjin: We don't need to re-serve the `rpc.Server` on every packet.
Just run;
```golang
go d.rpcS.ServeConn(d.clientConn)
```
within `d.Serve()`.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,160 @@
...
+		d.rpcC = rpc.NewClient(d.serverConn)
```
evanlinjin: We don't need to replace the `rpc.Client` on every packet.

evanlinjin (changes requested):

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,180 @@
...
+	conn       net.Conn
+	clientConn *branchConn
+	serverConn *branchConn
+	rpcS       *rpc.Server
+	rpcC       *rpc.Client
+	rpcUse     bool
+}
+
+// NewRPCDuplex initiates a new RPCDuplex struct
+func NewRPCDuplex(conn net.Conn, srv *rpc.Server, initiator bool, rpcUse bool) *RPCDuplex {
```
evanlinjin: Only change the production code to suit the test when absolutely necessary. In this case, having `rpcUse` is not okay.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	prefix := b[0]
+	size := binary.BigEndian.Uint16(b[1:3])
+
+	return prefix, size
+}
+
+// forward forwards one packet from Original conn to appropriate branchConn based on the packet's prefix
+func (d *RPCDuplex) forward() error {
```
evanlinjin, line 125: `readHeader` and `forward` can be combined to just `forward`.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,180 @@
...
+
+// readHeader reads the first three bytes of the data and returns the prefix and size of the packet
+func (d *RPCDuplex) readHeader() (byte, uint16) {
+
+	var b = make([]byte, 3)
+
+	// Read the first three byte of packet
+	// 1st byte is prefix and the 2nd and 3rd byte is the encoded size
+	_, err := d.conn.Read(b[:3])
+	if err != nil {
```
evanlinjin:
```golang
if _, err := d.conn.Read(b[:3]); err != nil {
    /* ... */
}
```

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,180 @@
...
+func (d *RPCDuplex) forward() error {
+
+	// Reads 1st byte of prefixed connection to determine which conn to forward to
+	prefix, size := d.readHeader()
+
+	// Reads packet with size 'size' from Original Conn into data
+	data := make([]byte, size)
+	// n, err := d.conn.Read(data)
+	_, err := d.conn.Read(data)
+	if err != nil {
```
evanlinjin: lines 138 & 139 can be combined.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,180 @@
...
+		if err == io.EOF {
+			return nil
+		}
+		return err
+	}
+
+	// Push data from original conn into branchConn's chan []byte
+	switch {
```
evanlinjin:
```golang
switch prefix {
    case d.serverConn.prefix:
        /* ... */
    case d.clientConn.prefix:
        /* ... */
}
```

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,180 @@
...
+	case prefix == d.serverConn.prefix:
+		// log.Println("serverConn", string(data[:n]))
```
evanlinjin: please remove commented code.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,180 @@
...
+		d.clientConn = &branchConn{Conn: conn, prefix: 0, readCh: clientCh}
+		d.serverConn = &branchConn{Conn: conn, prefix: 1, readCh: serverCh}
+	} else {
+		d.clientConn = &branchConn{Conn: conn, prefix: 1, readCh: clientCh}
+		d.serverConn = &branchConn{Conn: conn, prefix: 0, readCh: serverCh}
+	}
+	d.conn = conn
+	d.rpcS = srv
+
+	// Right now, the duplex can call RPC methods under the following condition:
```
evanlinjin: This large section of comments is not needed. Just looking at the code, it is clear what the logic is doing. There is not writing enough comments and there is over-doing comments.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+		log.Println("aDuplex:", r)
+
+		require.NoError(t, dB.Client().Call("RPC.Double", i, &r))
+		require.Equal(t, i*2, r)
+		log.Println("bDuplex:", r)
+	}
+}
+
+// TestbranchConn_Read reads data pushed in by
+// the Original connection
```
evanlinjin: These comments do not belong here...
And we need comments above each sub-test.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+func TestBranchConn_Read(t *testing.T) {
+
+	t.Run("successful_branchConn_read", func(t *testing.T) {
+
+		ch := make(chan []byte)
+		bc := &branchConn{prefix: 0, readCh: ch}
+		msg := []byte("foo")
+
+		go func() {
```
evanlinjin, line 444: Instead of a separate go-routine, initiate `ch` with a size of `1`; `ch := make(chan []byte, 1)`

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+
+	connA, connB := net.Pipe()
+	var ch chan []byte
+	defer connB.Close()
+
+	bc := &branchConn{Conn: connA, readCh: ch}
+
+	nCh := make(chan int, 1)
+	errCh := make(chan error, 1)
+	go func() {
```
evanlinjin:
```golang
payload := "foo"

done := make(chan struct)
defer close(done)

var n int
var err error
go func() {
    n, err = bc.Write([]byte(payload))
    done <- struct{}
}()

// other calls and tests ...

<-done
// test bc.Write results here...
```

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+		errCh <- err
+	}()
+
+	msg, err := ioutil.ReadAll(connB)
+
+	assert.Equal(t, 3, <-nCh)
+	require.NoError(t, <-errCh)
+	require.NoError(t, err)
+	assert.Equal(t, 6, len(msg))
+	assert.Equal(t, string([]byte("\x00\x00\x03foo")), string(msg))
```
evanlinjin: This is unclear. Have the following checks separately.
- Ensure `len(msg)` is equal to `len(payload) + 3`.
- Ensure `msg[0]` is the right prefix.
- Ensure `msg[1:3]` decodes to `len(payload)`.
- Ensure `msg[3:]` is equal to `payload`.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+// RPC is a receiver which we will use Register to publishes the receiver's methods in the DefaultServer.
+type RPC struct{}
+
+// RPC methods must look schematically like: func (t *T) MethodName(argType T1, replyType *T2) error
+func (RPC) Double(i int, reply *int) error {
+	*reply = i * 2
+	return nil
+}
+
+func TestRPCDuplex_Serve(t *testing.T) {
```
evanlinjin, line 397: Have a comment explaining the test.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+		require.NoError(t, err)
+		require.NoError(t, <-errCh)
+		assert.Equal(t, 5, n)
+		assert.Equal(t, string([]byte("hello")), string(b))
+	})
+}
+
+// TestRPCDuplex_Forward forwards one packet Original conn to branchConn
+// based on the packet's prefix
+func TestRPCDuplex_Forward(t *testing.T) {
```
evanlinjin, line 364: If the intention of this test is to ensure that the forwarding logic works and that the right packets from the original connection is forwarded to the right branched connection;
- We won't be needing `rpc.Server` input (as we won't be using it).
- We only need one `RPCDuplex`. We can just push self-defined packets to the other end of `net.Pipe()`.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+
+// TestReadHeader reads the first-three bytes of the data and
+// returns the prefix and size of the packet
+func TestRPCDuplex_ReadHeader(t *testing.T) {
+
+	// Case Tested: aDuplex's clientConn is the initiator and has
+	// a prefix of 0 and wrote a msg "foo" with size 3
+	// 1) prefix: Want(0) -- Got(0)
+	// 1) size: Want(3) -- Got(3)
+	t.Run("successfully_read_prefix_from_clientConn_to_serverConn", func(t *testing.T) {
```
evanlinjin: If we are just testing `readHeader`, we don't need a second `RPCDuplex`, nor do we need `rpc.Server` as input (`nil` is fine). These two sub-tests are very similar, we should have a loop going through different scenarios.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+
+			assert.Nil(connA.SetDeadline(time.Now().Add(time.Second * 10)))
+			assert.Nil(connB.SetDeadline(time.Now().Add(time.Second * 10)))
+
+			// Create two instances of RPCDuplex and rpcServer
+			rpcSrvA := rpc.NewServer()
+			rpcSrvB := rpc.NewServer()
+
+			aDuplex := NewRPCDuplex(connA, rpcSrvA, true, false)
+			bDuplex := NewRPCDuplex(connB, rpcSrvB, false, false)
```
evanlinjin: use nil for `rpc.Server` input as we are not using it.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+// in the appropriate branchConn with multiple test cases
+func TestNewRPCDuplex(t *testing.T) {
+	for _, tt := range tables {
+		t.Run(tt.description, func(t *testing.T) {
+
+			b := make([]byte, 256)
+			assert := assert.New(t)
+			expectedMsgCount := 10
+			chWrite := make(chan chReadWrite, expectedMsgCount)
+			chRead := make(chan chReadWrite, expectedMsgCount)
```
evanlinjin, line 276: `defer close(chWrite)`
`defer close(chRead)`

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+					msg := fmt.Sprintf(tt.msg, i)
+					n, err := bcWriter.Write([]byte(msg))
+					chWrite <- chReadWrite{i: i, n: n, err: err, msg: msg}
+				}
+				close(chWrite)
+			}()
+
+			// Read prefix and forward to the appropriate channel
+			errChA := make(chan error)
+			go func() { errChA <- aDuplex.Serve() }()
```
evanlinjin: Instead of calling server, we can loop through calls to `forward`.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+			close(errChA)
+
+			errChB := make(chan error)
```
evanlinjin: declare with size `1`

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,430 @@
...
+				for i := 0; i < expectedMsgCount; i++ {
+					msg := fmt.Sprintf(tt.msg, i)
+					n, err := bcWriter.Write([]byte(msg))
+					chWrite <- chReadWrite{i: i, n: n, err: err, msg: msg}
+				}
+				close(chWrite)
+			}()
+
+			// Read prefix and forward to the appropriate channel
+			errChA := make(chan error)
```
evanlinjin: declare with size 1

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,180 @@
...
+	"io"
+	"log"
+	"net"
+	"net/rpc"
+)
+
+const bufferSize = 8
+
+// branchConn will inherit the net.Conn interface.
+type branchConn struct {
```
evanlinjin, line 16: We need to have a `Close` function of `branchConn` perform the following;
- close the internal `readCh`.
- end all awaiting `branchConn.Read` calls.
- also close the sibling `branchConn` (e.g. closing `serverConn` should also close `clientConn`).
- end `RPCDuplex.Serve()`.

evanlinjin (changes requested):
Good work so far. Please review the following.

pkg/util/netutil/duplex.go
evanlinjin: Please implement this as follows:
- Have a member function: `RPCDuplex.closeDone()` under the protection of `sync.Once` that closes both of the `branchConn.readCh`s.
- `RPCDuplex.Close` should call `RPCDuplex.closeDone` (under `sync.Once` of course).
- `branchConn.Close` should call `RPCDuplex.Close`.
- When we call `branchConn.Read`, have `data, ok := <-bc.readCh` (and return `io.ErrClosedPipe` if `!ok`).

Then have the following additional tests:
- Closing any `branchConn` or `RPCDuplex` should close everything (a.k.a. `readCh` should panic when we push data to it with "write to closed channel" or something, and `RPCDuplex.Serve` should return).
- Closing any `branchConn` or `RPCDuplex` multiple times shouldn't result in panic (error is fine here).

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,435 @@
...
+			bc := &branchConn{Conn: connA, readCh: ch}
+			payload := "foo"
+
+			done := make(chan struct{})
+			defer close(done)
+
+			go func() {
+				_, err = bc.Write([]byte(payload))
+				bc.Conn.Close()
+				done <- struct{}{}
```
evanlinjin, line 126: We can just have `close(done)` here and remove line 120.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,435 @@
...
+		var b []byte
+		n, err := bc.Read(b)
+
+		require.NoError(t, err)
+		assert.Equal(t, 0, n)
+		assert.Equal(t, []byte(nil), b)
+	})
+}
+
+var tablesWrite = []struct {
```
evanlinjin: Declare this within `TestBranchConn_Write`. This shouldn't be a public variable.

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,435 @@
...
+	lenMsg      int
+	prefix      byte
+	lenPayload  uint16
+	expectedMsg []byte
+}{
+	// TestbranchConn_Writes writes len(p) bytes from p to
+	// data stream and appends it with a 1 byte prefix and
+	// 2 byte encoded length of the packet. Thus, length
+	// of message should equal to length of payload + 3
+	{description: "len(msg)_equal_len(payload)+3",
```
evanlinjin: Aren't these all the same tests?
These comments seem very unnecessary. Please have comments on top of `TestBranchConn_Write` that describes what is being testing with each case, and not describe each case.

pkg/util/netutil/duplex_test.go
evanlinjin: The other option is to use `sync.WaitGroup`

evanlinjin (changes requested):

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,169 @@
...
+	"log"
+	"net"
+	"net/rpc"
+	"sync"
+)
+
+const bufferSize = 8
+
+// closeCh ensures that branchConn's readCh channel is closed once
+var closeCh sync.Once
```
evanlinjin: We can only call a function under `closeCh` once. Hence, putting this here results in `RPCDuplex.Close` only ever being called once (but for all instances of `RPCDuplex`).

In other words, after we close one instance of `RPCDuplex`, future instances of `RPCDuplex` cannot be closed. This is not the right behavior.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,169 @@
...
+	// Reads in data from chan []byte pushed from Original Conn
+	data, ok := <-bc.readCh
+	if !ok {
+		return 0, io.ErrClosedPipe
+	}
+
+	// Compare length of data with b to check if a longer buffer is required
+	if len(data) > len(b) {
+		err = io.ErrShortBuffer
+		return n, err
```
evanlinjin: Have `return 0, io.ErrShortBuffer` for some consistency.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,169 @@
...
+
+// Close calls Duplex.Close to close internal readCh for both branchConn
+func (bc *branchConn) Close() error {
+	return bc.duplex.Close()
+}
+
+// Read reads in data from original conn through branchConn's chan [].
+// If read []byte is smaller than length of packet, io.ErrShortBuffer
+// is raised.
+func (bc *branchConn) Read(b []byte) (n int, err error) {
```
evanlinjin: `func (bc *branchConn) Read(b []byte) (int, error) {`

pkg/util/netutil/duplex.go
evanlinjin: Also `closeCh` is a bad name. This is not a channel.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Close calls closeDone which close both branchConn's readCh
+func (d *RPCDuplex) Close() error {
+
+	closeCh.Do(func() {
+		d.closeDone()
+	})
+
+	return errors.New("branchConn closed")
```
evanlinjin: Always returning an error is not the correct behavior. We should only return error when already closed.

pkg/util/netutil/duplex.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+// Serve is a blocking function that serves the RPC server and runs the event loop that forwards data to branchConns.
+func (d *RPCDuplex) Serve() error {
+
+	go d.rpcS.ServeConn(d.serverConn)
+
+	for {
+		err := d.forward()
```
evanlinjin:
```golang
if err := d.forward(); err != nil {
    if err == io.EOF {
        return nil
    }
    return err
}
```

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,495 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBranchConn_Read(t *testing.T) {
+
+	// Read reads in data from original conn through branchConn's readCh which is
+	// a type of chan [].
```
evanlinjin: Please describe the test (aka the expected behavior). For example;
```
// It is expected that a 'branchConn' can Read data that is pushed in it's 'readCh' successfully.
```

evanlinjin: And what is `a type of chan []`?

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,495 @@
...
+		b := make([]byte, 3)
+		n, err := bc.Read(b)
+
+		require.NoError(t, err)
+		assert.Equal(t, 3, n)
+		assert.Equal(t, []byte("foo"), b)
+	})
+
+	// Read reads in empty data from original conn through branchConn's readCh which is
+	// a type of chan []. It should return 0 length read and b should be nil
```
evanlinjin:
```
// It is expected that 'branchConn' handles empty packets (read via 'readCh') correctly.
// In such a scenario, 'branchConn.Read()' should return 'err == nil' and 'n == 0'.

File: pkg/util/netutil/duplex_test.go [hunk 50]
```diff
@@ -0,0 +1,495 @@
...
+		initiatorB:  true,
+		whoWrites:   "bDuplex",
+		branchConn:  "serverConn",
+		expectedMsg: "helloworld. Skycoin is best coin!%d",
+	},
+}
+
+// Test sending multiple messages in a single connection and receiving them
+// in the appropriate branchConn with multiple test cases
+func TestNewRPCDuplex(t *testing.T) {
```
evanlinjin, line 268: This is a very complicated test. Please provide a better description.
Also wondering whether we need this and whether it can be simplified...

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,495 @@
...
+// TestRPCDuplex_Forward forwards one packet Original conn to branchConn
+// based on the packet's prefix
+func TestRPCDuplex_Forward(t *testing.T) {
+
+	connA, connB := net.Pipe()
+	defer connB.Close()
+
+	bDuplex := NewRPCDuplex(connB, nil, false)
+
+	msg := []byte("\x00\x00\x03foo")
```
evanlinjin:
```golang
p := append([]byte{0, 0, 3}, []byte("foo")...)
```

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,495 @@
...
+			// Assert error from msgReader.Forward()
+			assert.Nil(<-errChA)
+			assert.Nil(<-errChB)
+
+		})
+	}
+}
+
+// TestRPCDuplex_Forward forwards one packet Original conn to branchConn
+// based on the packet's prefix
```
evanlinjin: What is the correct behavior? What are we testing?
```
Given ... it is expected that ....
```

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,495 @@
...
+
+// RPC methods must look schematically like: func (t *T) MethodName(argType T1, replyType *T2) error
+func (RPC) Double(i int, reply *int) error {
+	*reply = i * 2
+	return nil
+}
+
+// TestRPCDuplex_Serve test communication between two RPC Duplex. When both Duplex
+// is created and served, both duplex's client should be able to successfully call
+// the rpc method 'double' from rpc.server
```
evanlinjin: `from the remote's rpc.Server`

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,495 @@
...
+	assert.Nil(t, <-errChA)
+	assert.Nil(t, <-errChB)
+
+}
+
+func TestRPCDuplex_Close(t *testing.T) {
+
+	// Closing RPCDuplex should close both branchConn's readCh; therefore, it
+	// should panic with "write to closed channel" when data is pushed to it
+	t.Run("readCh_panic_when_write_read_to_closed_channel", func(t *testing.T) {
```
evanlinjin: `branchConns_should_close_when_RPCDuplex_closes`

pkg/util/netutil/duplex_test.go
```diff
@@ -0,0 +1,495 @@
...
+	close(errChB)
+	assert.Nil(t, <-errChA)
+	assert.Nil(t, <-errChB)
+
+}
+
+func TestRPCDuplex_Close(t *testing.T) {
+
+	// Closing RPCDuplex should close both branchConn's readCh; therefore, it
+	// should panic with "write to closed channel" when data is pushed to it
```
evanlinjin: This should describe WHAT we are testing, not HOW we are testing.
E.g. Using `pushing to closed chan` is used to test whether `branchConn` is closed. This is a HOW.

evanlinjin (changes requested):

pkg/util/netutil/duplex_test.go
```diff
@@ -307,13 +303,10 @@ func TestNewRPCDuplex(t *testing.T) {
...
-					err := aDuplex.forward()
-					switch err {
-					case nil:
-						continue
-					case io.EOF:
-						errChA <- err
-					default:
+					if err := aDuplex.forward(); err != nil {
+						if err == io.EOF {
+							errChA <- nil
```
evanlinjin: This results in two variables being pushed into `errChA` channel. A more suitable solution is:
```golang
if err == io.EOF {
    err = nil
}
```

pkg/util/netutil/duplex_test.go
```diff
@@ -323,13 +316,10 @@ func TestNewRPCDuplex(t *testing.T) {
...
-					err := bDuplex.forward()
-					switch err {
-					case nil:
-						continue
-					case io.EOF:
-						errChB <- err
-					default:
+					if err := bDuplex.forward(); err != nil {
+						if err == io.EOF {
+							errChB <- nil
```
evanlinjin: Same here!

pkg/util/netutil/duplex_test.go
```diff
@@ -18,8 +17,8 @@ import (
 
 func TestBranchConn_Read(t *testing.T) {
 
-	// Read reads in data from original conn through branchConn's readCh which is
-	// a type of chan [].
+	// It is expected that a 'branchConn' can Read data that is pushed in it's 'readCh' of type chan []byte successfully.
```
evanlinjin: please remove `of type chan []byte`. We can see that in the code.

evanlinjin (changes requested):

Issue #343, opened 2019-05-09
Handshake synchronyzation


ivcosla:
Concern raised by @nkryuchkov 

https://github.com/skycoin/skywire/blob/mainnet/pkg/transport/handshake.go#L48 until 56 those operations might need to be guarded by mutexes to prevent a wrong order between write and read operations, leading to chacha bug appearing when creating a transport.

ivcosla:
The following log is the result of an intermediary node restart cycle of more than 100 times in which it reconnects with nodes A and B, while at the same time messages are sent over chat app from node A to node C and back. 

Data races regarding to handshakes seem to disappear (#346 ) with this PR. There are however chacha errors, note that all of them are the same:

`[2019-05-09T21:05:06+02:00] WARN [trmanager]: Failed to re-establish transport: read: chacha20poly1305: message authentication failed`


[logB.txt](https://github.com/skycoin/skywire/files/3163613/logB.txt)

ivcosla:
If node B is not restarted, thus making route between nodes A and C stable chacha encryption error doesn't seems to appear while sending messages.

PR #344, opened 2019-05-09
[WIP] added mux as var on handshake initiator

ivcosla:
Attempt to #343


evanlinjin:
@ivcosla is this fix still relevant?

PR #349, opened 2019-05-12
BUG: Ensure modules using internal/httpauth retries with correct nonce after 401. 

### LINKED ISSUE #345: BUG: Ensure modules using internal/httpauth retries with correct nonce after 401.
Log extract:
```
╰─$ go run ./cmd/skywire-node swA.json                                                                                                                                        1 ↵
[2019-05-09T23:46:38+12:00] INFO [pathutil]: using args[0] as config path: swA.json
[2019-05-09T23:46:56+12:00] INFO [messenger]: Opened new channel local ID 3, remote ID 1 with 0348c941c5015a05c455ff238af2e57fb8f914c399aab604e9abb5b32b91a4c1fe
[2019-05-09T23:46:56+12:00] WARN [trmanager]: Failed to accept connection: entry set: status: 401, error: {"error":{"message":"SW-Nonce does not match","code":401}}
```

The client after receiving `{"error":{"message":"SW-Nonce does not match","code":401}}`, must:
1. Grab correct nonce via `(*httpauth.Client).Nonce()`.
2. Retry.

nkryuchkov:
Closes #345.

In this PR, `internal/httpauth` requests a new nonce and retries HTTP request if a nonce validation error is returned.

The new behavior may be tested running the following commands:
```
go test ./internal/httpauth -run TestClient_BadNonce
```

### DIFF
--- a/internal/httpauth/client.go
+++ b/internal/httpauth/client.go
@@ -45,7 +45,7 @@ type Client struct {
 	key    cipher.PubKey
 	sec    cipher.SecKey
 	Addr   string // sanitized address of the client, which may differ from addr used in NewClient
-	nonce  uint64
+	nonce  uint64 // has to be handled with the atomic package at all time
 }
 
 // NewClient creates a new client setting a public key to the client to be used for Auth.
@@ -164,7 +164,7 @@ func (c *Client) doRequest(req *http.Request, body []byte) (*http.Response, erro
 }
 
 func (c *Client) getCurrentNonce() Nonce {
-	return Nonce(c.nonce)
+	return Nonce(atomic.LoadUint64(&c.nonce))
 }
 
 func (c *Client) incrementNonce() {
--- a/internal/httpauth/client.go
+++ b/internal/httpauth/client.go
@@ -13,8 +13,14 @@ import (
 	"strconv"
 	"strings"
 	"sync/atomic"
+	"time"
 
 	"github.com/skycoin/skywire/pkg/cipher"
+)
+
+const (
+	retryInterval            = 100 * time.Millisecond
+	invalidNonceErrorMessage = "SW-Nonce does not match"
 )
 
 // NextNonceResponse represents a ServeHTTP response for json encoding
@@ -71,11 +77,6 @@ func NewClient(ctx context.Context, addr string, key cipher.PubKey, sec cipher.S
 // Do performs a new authenticated Request and returns the response. Internally, if the request was
 // successful nonce is incremented
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
-	req.Header.Add("SW-Public", c.key.Hex())
-
-	// use nonce, later, if no err from req update such nonce
-	req.Header.Add("SW-Nonce", strconv.FormatUint(uint64(c.getCurrentNonce()), 10))
-
 	body := make([]byte, 0)
 	if req.ContentLength != 0 {
 		auxBody, err := ioutil.ReadAll(req.Body)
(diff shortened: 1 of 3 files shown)


internal/httpauth/client.go
```diff
@@ -86,16 +87,42 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
...
+
+		// use nonce, later, if no err from req update such nonce
+		req.Header.Set("SW-Nonce", strconv.FormatUint(uint64(nonce), 10))
+		req.Header.Set("SW-Sig", sign.Hex())
+		req.Header.Set("SW-Public", c.key.Hex())
 
-	res, err := c.client.Do(req)
-	if err != nil {
-		return nil, err
+		res, err = c.client.Do(req)
```
ayuryshev: In "theory" recursive calls could continue now infinitely.
IMO it could be be improved with recursion call limit.
With `return nil,err` on overruns.

internal/httpauth/client.go
```diff
@@ -86,16 +87,42 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
...
+		}
+
+		nonce, err = c.Nonce(context.Background(), c.key)
+		if err != nil {
+			return nil, err
+		}
+		c.SetNonce(nonce)
+
+		res.Body.Close()
+		time.Sleep(retryInterval)
```
ayuryshev: `retryInterval` would be better if it was a settings value IMO

ayuryshev (changes requested):
I think it could be improved with introducing two configuration values:
- timeout between sequential `Client.Do`  requests for nonce
- recursion limit for `Client.Do` requests

internal/httpauth/client.go
ivcosla: agree

internal/httpauth/client.go
ivcosla: About this I don't feel it's a parameter that the user is going to be concerned about, I feel it's fine as a constant in package.

ivcosla (approved):
Good job.

internal/httpauth/client.go
nkryuchkov: Thanks, I think it's a good idea. Added a commit with the implementation of max retries threshold as well as a test for it.

internal/httpauth/client.go
evanlinjin: Why do we need a retry parameter? It is not a network error.

internal/httpauth/client.go
```diff
@@ -86,16 +88,46 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
...
 		body = auxBody
 	}
-	sign, err := Sign(body, c.getCurrentNonce(), c.sec)
-	if err != nil {
-		return nil, err
-	}
 
-	req.Header.Add("SW-Sig", sign.Hex())
+	var res *http.Response
+	for retriesCount := 0; retriesCount < maxRetries; retriesCount++ {
```
evanlinjin: As we are not dealing with a network error, I suggest the following changes:
1. One retry is plenty. If nonce is not synced, we only need to sync once afterwards.
2. We will not need a timeout or `retryInterval`.

evanlinjin: Instead of a `for` loop, I would group lines `94-121` into a closure (excluding 110-112). And do `c.Nonce()` + `c.SetNonce()` + `call-closure-again` on error.

internal/httpauth/client.go
```diff
@@ -144,6 +176,26 @@ func (c *Client) incrementNonce() {
 	atomic.AddUint64(&c.nonce, 1)
 }
 
+func (c *Client) isNonceValid(res *http.Response) (bool, error) {
```
evanlinjin: 1. Please leave a comment on this function.
2. We aren't using `c *Client`, please have it removed.

internal/httpauth/client_test.go
```diff
@@ -15,34 +17,140 @@ import (
...
 	res.Body.Close()
-	assert.Equal(t, []byte("Hello, client\n"), b)
+	assert.Equal(t, []byte(payload), b)
 	assert.Equal(t, uint64(2), c.nonce)
+
+	headers := <-headerCh
+	checkResp(t, headers, b, pk, 1)
+}
+
+func TestBadNonce(t *testing.T) {
```
evanlinjin: Please rename to `TestClient_BadNonce` and add a comment summarizing the test.

internal/httpauth/client_test.go
```diff
@@ -15,34 +17,140 @@ import (
...
+	b, err := ioutil.ReadAll(res.Body)
+	require.NoError(t, err)
+	res.Body.Close()
+	assert.Equal(t, uint64(2), c.nonce)
+
+	headers := <-headerCh
+	checkResp(t, headers, b, pk, 1)
+}
+
+func TestTooManyRetries(t *testing.T) {
```
evanlinjin: Please rename to `TestClient_TooManyRetries` and add a comment summarizing the test.

evanlinjin (changes requested):
Good work! Please review my comments.

internal/httpauth/client_test.go
nkryuchkov: Since we do not need to retry multiple times anymore, the test is removed.

nkryuchkov:
@evanlinjin The requested changes are made. Could you review the last changes, please?

internal/httpauth/client.go
```diff
@@ -136,6 +148,21 @@ func (c *Client) SetNonce(n Nonce) {
...
+
+	// use nonce, later, if no err from req update such nonce
+	req.Header.Set("SW-Nonce", strconv.FormatUint(uint64(nonce), 10))
+	req.Header.Set("SW-Sig", sign.Hex())
+	req.Header.Set("SW-Public", c.key.Hex())
+
+	return c.client.Do(req)
+}
+
 func (c *Client) getCurrentNonce() Nonce {
```
evanlinjin, line 171: Operations to `c.nonce` should be atomic. Please use [`atomic.LoadUint64`](https://golang.org/pkg/sync/atomic/#LoadUint64). Also add a comment to line 48, that `c.nonce` is to be handled with the atomic package at all times.

evanlinjin (changes requested):
Looks good. Just one more thing.

ayuryshev (approved):

internal/httpauth/client.go
nkryuchkov: Fixed

evanlinjin (approved):

Issue #350, opened 2019-05-15
Error on updating entry


ivcosla:
Related to https://github.com/skycoin/skywire/issues/347, before node B fails during startup process and messages are not delivered between A and C this log appears on B:

```bash
[2019-05-14T13:51:51+02:00] INFO [trmanager]: Failed to re-dial Transport 38254102-9309-5c24-85c0-d3098eb81256: link failure: entry update failure: status code: 422. message: entry validation error: sequence field of new entry is not sequence of old entry + 1
```
For messaging-discovery client, it might happen that latest update has not been applied yet on the server side, thus when a new one happens sequence might differ:
 - [x] Retry for newest sequence on `UpdateEntry`, getting latest sequence from server
 - [x] Write test for retrial

evanlinjin:
Related to https://github.com/skycoin/skywire/pull/349

PR #351, opened 2019-05-15
added retry on sequence error for update

ivcosla:


ivcosla:
The bug seems fixed, however we might want to make some changes to how to deal with retry.

ivcosla:
Closes #350

jdknives:
This is ready for review? @ivcosla

ivcosla:
yes

pkg/messaging-discovery/client/client.go
```diff
@@ -108,27 +109,47 @@ func (c *httpClient) SetEntry(ctx context.Context, e *Entry) error {
...
 	if err != nil {
 		return err
 	}
 
-	err = c.SetEntry(ctx, e)
-	if err != nil {
-		e.Sequence--
+	for {
+		err = c.SetEntry(ctx, e)
+		if err != nil {
```
nkryuchkov, line 127: I would check if `err == nil` here and `return nil` in this case. IMO this would make to code clearer and allow us to remove one level of indentation along with `continue` statement.

Darkren:
Agree with @nkryuchkov on stated above. Otherwise looks fine to me

pkg/messaging-discovery/client/testing.go
```diff
@@ -91,7 +91,30 @@ func (m *mockClient) UpdateEntry(ctx context.Context, sk cipher.SecKey, e *Entry
 		return err
 	}
 
-	return m.SetEntry(ctx, e)
+	for {
+		err = m.SetEntry(ctx, e)
+		if err != nil {
```
nkryuchkov, line 92: Could you please invert the if statement and remove the `continue` statement below too?

ayuryshev (approved):

nkryuchkov (approved):

pkg/messaging-discovery/client/client.go
evanlinjin: Also consider moving `e.Sign` to the top of the `for` loop. This way, we can remove 124-127.

evanlinjin (changes requested):
Looks good. Some requests.

pkg/messaging-discovery/client/client.go
```diff
@@ -108,27 +109,46 @@ func (c *httpClient) SetEntry(ctx context.Context, e *Entry) error {
...
 	if err != nil {
 		return err
 	}
 
-	err = c.SetEntry(ctx, e)
-	if err != nil {
-		e.Sequence--
+	for {
+		err = c.SetEntry(ctx, e)
+		if err == nil {
```
evanlinjin, line 131: Also I would have `if err != nil` here and have error-handling code within that. This way it explicitly states; "This is error-handling code". But this is a matter of personal-taste and you can choose otherwise :P

Issue #365, opened 2019-05-23
Cleanup: Elimination of closing channels and copying data in Manager.Observe


ayuryshev:
# Analysis

```golang
func (tm *Manager) Observe() (accept <-chan *ManagedTransport, dial <-chan *ManagedTransport) {
	dialCh := make(chan *ManagedTransport)
	acceptCh := make(chan *ManagedTransport)
	go func() {
		for {
			select {
			case <-tm.doneChan:
				close(dialCh)
				close(acceptCh)
				return
			case tr := <-tm.acceptedTrChan:
				acceptCh <- tr
			case tr := <-tm.dialedTrChan:
				dialCh <- tr
			}
		}
	}()
	return tm.AcceptedTrChan, tm.DialedTrChan
}
```
Those `close(dialCh)` and `close(acceptCh)` are possible cause of `panic: send on closed channel` in  #354 

Copying data from `tm.acceptedTrChan` and `tm.dialedTrChan` looks dubious too. It's a possible source of more subtle errors.

# TODO 

Eliminate or justify existence of `close` instructions and copying data.

1. Remove dubious instructions
2. Test it
3. Publish results

gz-c:
Is this code snippet the actual code?

ayuryshev:
It was eliminated in https://github.com/skycoin/skywire/pull/366
No reason for existence of this lines was not found.

jdknives:
@gz-c yes. That is actual code in `mainnet`.

ayuryshev:
Done

Issue #354, opened 2019-05-16
Data races on exit of Skywire-nodes


ayuryshev:
# How to reproduce bug

1. Start 3 skywire-nodes: 
- NodeA with thereallssh app
- intermediary NodeB without apps
- NodeC with therealssh-client app
2. Add transports from NodeA and NodeC to NodeB
3. Start ssh communication. E.g. `therealssh-cli $PK_A "echo A"` (or `cat` a file, `ls`, `tree`, etc)
4. Kill nodes during proccess

With high probability data races will be on one or two of nodes.

Sometimes data race is followed with `panic: send on closed channel` and sometimes not.

Attached logs:
- nodeB.log and nodeC.log - are from one run
- nodeA.log - from another

[nodeC.log](https://github.com/skycoin/skywire/files/3186281/nodeC.log)
[nodeB.log](https://github.com/skycoin/skywire/files/3186282/nodeB.log)
[nodeA.log](https://github.com/skycoin/skywire/files/3186283/nodeA.log)

gz-c:
How do i configure NodeA, NodeB, NodeC?

jdknives:
Closed by #361

PR #366, opened 2019-05-23
Cleanup: Eliminate or justify Manager.Observe

ayuryshev:
1. `Manager.Observe` eliminated
2. All tests passed in skywire and skywire-services

Closes: #365 
Partially relates to: #354

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -78,9 +78,10 @@ func New(config *Config) *Router {
 
 // Serve starts transport listening loop.
 func (r *Router) Serve(ctx context.Context) error {
-	acceptCh, dialCh := r.tm.Observe()
+	// acceptCh, dialCh := r.tm.Observe()
+
 	go func() {
-		for tr := range acceptCh {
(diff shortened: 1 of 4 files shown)


evanlinjin (approved):
I'm not sure how this is going to fix any bugs. But it's simplifications and doesn't change behaviour so I will approve and merge.

However it doesn't change what we have discussed in private chat:

**@ayuryshev on Discord:**
```
4.  Can we merge dialCh  and acceptCh and process them in single for-loop ?
Can we put everything in single for-loop with select ?
If it's possible - it will eliminate all possibilities for races in principle
And it coud tried quikly - it's not a major rewrite
What do you think?
```
**@evanlinjin on Discord:**
```
I see
We still need to differentiate accepted Transports from dialed transports
I think lets not touch transport manager, and just use a select statement in router.Serve
For tm.Observe, we can remove dialCh and acceptCh; they are useless.
```

ivcosla (approved):
I also aprove. Not solving anything, but cleaning code.

Issue #346, opened 2019-05-09
Data race on handshake


ivcosla:
[noise-data-race.txt](https://github.com/skycoin/skywire/files/3162724/noise-data-race.txt)

Might be related to #343

evanlinjin:
@ivcosla isn't this data race fixed in #337?

PR #337, opened 2019-05-08
Fix messaging system panic #336.

evanlinjin:
closes #336

PR #369, opened 2019-05-25
Attempt to fix race in Channel/Link shutdown

gz-c:
Changes the channel close logic to (hopefully) eliminate race conditions.

There is a single `done` chan that can be closed to signal closing.
To check if the channel has been closed, use a non-blocking `select` on the `done` chan.

`sync.Once` is used to ensure the channel is closed only once (since closing more than once would panic the application).

There are two sources of "closing" the channel:

- The transport owner calls `Close()`
- The client receives a `FrameTypeChannelClosed` message

The `waitChan` still seems inappropriate but I've left it there.

### DIFF
--- a/pkg/messaging/channel.go
+++ b/pkg/messaging/channel.go
@@ -153,18 +153,20 @@ func (mCh *msgChannel) Write(p []byte) (n int, err error) {
 	}
 }
 
-func (mCh *msgChannel) RequestClose() error {
+func (mCh *msgChannel) Close() error {
 	select {
 	case <-mCh.doneChan:
 		return ErrChannelClosed
 	default:
 	}
 
-	if _, err := mCh.link.SendCloseChannel(mCh.ID()); err != nil {
-		return err
+	if mCh.close() {
+		if _, err := mCh.link.SendCloseChannel(mCh.ID()); err != nil {
+			return err
+		}
 	}
 
-	mCh.close()
+	return nil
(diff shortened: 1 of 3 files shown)


evanlinjin:
From my understanding, this is an elegant solution to bug #346 (which is already fixed by PR #337). However, from my tests, it does not fix #354 

But because it cleans up the code-base, I think it can be merged after the test passes... (Something that confuses me too and am trying to fix atm).

evanlinjin (commented):
Looks good. However, I'll try fix the failing test before merging.

ivcosla:
I have also tested the branch and can confirm it doesn't fix the Data Race, howhever I also think that once the test is passing (maybe it has to do with `FrameTypeChannelClosed` handling) it's worth to merge it since it's a better solution than the previous one.

gz-c:
I took a guess at how to handle the `waitChan`; this is the next design aspect I would consider for revision.  


Also: the `bool` return value from `close()` isn't used anymore. I had it in place temporarily, to replicate the old behavior where the `clientLink.chans.remove(channelID)` was only called if the channel was closed by `close()`, which I've removed.  I don't know if that's correct, and I'd be concerned about a race condition where the channelID had been recycled and the `remove(channelID)` call affects a different channel instance.  One way to avoid that problem is to have a separate step where the channels are iterated and purged if they've been closed.

PR #372, opened 2019-05-26
Removal of transport.TCPFactory

ayuryshev:
```bash
rm pkg/transport/tcp_transpor*.go
make test
cd ../skywire-services
make dep ## with uncommented replace directive
make test
```

Integration tests obviously passes too.

Now `transport.Factory` == `messaging.Client` (excluding tests)


jdknives:
What is the purpose of this PR?

evanlinjin:
`TCPTransport` is a feature that Brandon specifically asked for. We may have not integrated it in the code-base yet, but it is needed.

ayuryshev:
Purpose - remove code without purpose. 
Every unneeded LOC - is potential source of bugs.
At least it will make `make test` slightly faster.

PR #375, opened 2019-05-27
[WIP] Bug/multiple transports

### LINKED ISSUE #364: Multiple transport creation
As stated in https://github.com/skycoin/skywire/issues/355, by looking at the logs we can see multiple `closing transport` log messages, those ones are generated here:

https://github.com/skycoin/skywire/blob/mainnet/pkg/transport/manager.go#L325

https://github.com/skycoin/skywire/blob/mainnet/pkg/transport/manager.go#L399

This comes a long way already, since we changed the way that transports are spawned in this PR https://github.com/skycoin/skywire/pull/252.

Before that, multiple transports were possible between two nodes, so in order to differenciate between them we had every transport identified by an uuid. Now there is only one possible transport between each pair of nodes, and will always have the same ID.

As for now, not all the code has been adapted from the previous implementation and we can see that handler go-routines linked before are still spawned when a transport is Dialed or Accepted.

Here I see two options:

1) Keep a map of remote PKs in transport manager, meaning existent transports to that remote. Then, don't dial a new Transport if there is already one. We need to make sure that transports are closed on failures.

2) Keep working with IDs, which are not available until the transport is created, but then, if such transport already exists drop it's handling inmediately and return an error. Which will also need to be handled gracefully by callers.

In both cases it is important that re-dial happens on error by: removing the previous transport, then dialing again.

ivcosla:
Attempts to fix #364

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -326,7 +326,7 @@ func (r *Router) requestLoop(appConn *app.Protocol, raddr *app.Addr) (*app.Addr,
 	if err != nil {
 		return nil, err
 	}
-	defer tr.Close()
+	defer r.tm.DeleteTransport(tr.ID)
 
 	if err := setup.CreateLoop(proto, l); err != nil {
 		return nil, fmt.Errorf("route setup: %s", err)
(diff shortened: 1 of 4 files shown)


jdknives:
Does this need to be scrapped?

ayuryshev:
Yes. It was resolved in https://github.com/skycoin/skywire/pull/386

Issue #327, opened 2019-05-03
Clarify shutdown logic.


evanlinjin:
## Background
Due to the way `router.Serve()` is defined (with event loops in go-routines that start event loops in more go-routines), shutdown logic is unclear and hard to follow. The following steps are to clarify these processes.

Extracts from conversations on Discord:

> tpm.Serve() should be move out. Yes.
calling router.New() has TransportManager as input, one should have run tpm.Serve() beforehand.
We should have RoutingTableManager as an input for router.New(), instead of RoutingTable. rtm.Run() should also be run beforehand.

> From my perspective, we can just ensure that we shutdown the components in the right order.
Each component has a .Close() function.
In skywire/cmd/skywire-node/commands/root.go we should call .Close() in the right order.
In regards to how we deal with panics... from my understanding, that will be a topic of discussion.
I'm going to create a channel in the main group for this conversation.

jdknives:
Closed by #376

PR #376, opened 2019-05-28
Feature: reorganize ManagedTransport concurrency

ayuryshev:
There are 7 changes in this PR.

This PR includes changes in PR #380  - that solved issue with failing first message.

### 2. `Manager.entries` now unique and `ReconnectTransports` now thread-safe

Changes:
    
1. Manager.entries now map[Entry]struct{} (was slice)
2. Work with with them in ReconnectTransports now locked by mutex
    properly
    
Tested both in skywire, skywire-services and by integration tests

### 3. `router.Serve`, `setup.Node`: serving accepted and dialled transports in the same loop

1. `ManagedTransport`: new field `Accepted bool` - to distinguish dialled
and accepted transports
2. `Manager`: it's only one channel now for both accepted and dialled
transports
3. `Router.Serve`: it's only one loop for both accepted and dialled
transports
4. `setup.Node` changes in the same way
   
### 4. More strict typing in PacketRouter, Router, mockRouter

Removed overgeneralization of `ManagedTransport` as `Transport` where it makes no sense.

### 5. Manager.manageTransport

Goroutines in Manager.acceptTransport and Manager.createTransport are    extracted into method  Manager.manageTransport

### 7. Incorrect shutdown: redialling on close

Solved.

Closes: #327
Relates to: #364


pkg/transport/manager.go
```diff
@@ -259,31 +255,49 @@ func (tm *Manager) DeleteTransport(id uuid.UUID) error {
...
 		tr.Close()
 	}
-	tm.transports = make(map[uuid.UUID]*ManagedTransport)
 	tm.mu.Unlock()
 
 	if _, err := tm.config.DiscoveryClient.UpdateStatuses(context.Background(), statuses...); err != nil {
 		tm.Logger.Warnf("Failed to change transport status: %s", err)
 	}
 
+	for _, f := range tm.factories {
```
ayuryshev, line 277: This was the first cause of redialling  during shutdown

pkg/transport/manager.go
```diff
@@ -416,10 +379,43 @@ func (tm *Manager) walkEntries(walkFunc func(*Entry) bool) *Entry {
...
 }
 
+func (tm *Manager) manageTransport(ctx context.Context, managedTr *ManagedTransport, factory Factory, remote cipher.PubKey, public bool, accepted bool) {
+	select {
+	case <-managedTr.doneChan:
+		tm.Logger.Infof("Transport %s closed", managedTr.ID)
+		return
+	case err := <-managedTr.errChan:
+		managedTr.mu.Lock()
+		isClosing := managedTr.isClosing
```
ayuryshev: And that prevents redialling on errors during shutdown

pkg/transport/manager.go
```diff
@@ -368,33 +346,18 @@ func (tm *Manager) acceptTransport(ctx context.Context, factory Factory) (*Manag
...
 	}
 	tm.mu.Unlock()
 
-	// go func(managedTr *ManagedTransport, tm *Manager) {
-	go func() {
-		select {
-		case <-managedTr.doneChan:
-			tm.Logger.Infof("Transport %s closed", managedTr.ID)
-			return
-		case <-tm.doneChan:
```
ayuryshev, line 388: And this was very subtle bug. Must be more evident with messaging tests with more than 2 skychat-nodes

evanlinjin (approved):

pkg/transport/manager.go
```diff
@@ -54,23 +53,18 @@ func NewManager(config *ManagerConfig, factories ...Factory) (*Manager, error) {
...
-		entries:        mEntries,
-		AcceptedTrChan: make(chan *ManagedTransport, 10),
-		DialedTrChan:   make(chan *ManagedTransport, 10),
-		doneChan:       make(chan struct{}),
+		Logger:     logging.MustGetLogger("trmanager"),
+		config:     config,
+		factories:  fMap,
+		transports: make(map[uuid.UUID]*ManagedTransport),
+		entries:    mEntries,
+		// AcceptedTrChan: make(chan *ManagedTransport, 10),
```
ivcosla: Remove this before merging

ivcosla (commented):
Is looking good for me, however please wait for integration test input before merging it.

pkg/transport/manager.go
ayuryshev: Done

ivcosla:
Overall is looking better than mainnet. Chat app restart for itnermediary node is quite reliable.

At first I had trouble running the PR because after the first run I restarted nodes and they all hanged on startup. Cleaning redis didn't help, the problem had to do with routing.db, however have not been able to reproduce this error again.

SSH is still working well, though we still have the problem where server and client apps hang after node shutdown. There is some problem with the integration test for SSH parsing arguments (cannot parse `echo`), but this is also present in mainnet at the moment, guess it was introduced during renaming.

Very good job!

ivcosla (approved):

Issue #355, opened 2019-05-16
Therealssh hangs after chacha20poly1305 error


ayuryshev:
# How to reproduce bug

1. Start 3 skywire-nodes: 
- NodeA with thereallssh app
- intermediary NodeB without apps
- NodeC with therealssh-client app
2. Add transports from NodeA and NodeC to NodeB
3. Start ssh communication. E.g. `therealssh-cli $PK_A "echo A"` (or `cat` a file, `ls`, `tree`, etc)
4. Increase amount of data. On some point of time NodeC logs:

```
[2019-05-16T12:19:03+03:00] WARN [router]: Stopped serving Transport: chacha20poly1305: message authentication failed
[2019-05-16T12:19:03+03:00] INFO [trmanager]: Transport 38254102-9309-5c24-85c0-d3098eb81256 failed with error: chacha20poly1305: message authentication failed. Re-dialing...
[2019-05-16T12:19:03+03:00] INFO [trmanager]: De-registered transport 38254102-9309-5c24-85c0-d3098eb81256
```

5. Stop NodeC. On exit NodeC logs:
```

[2019-05-16T12:24:21+03:00] INFO [router]: App request with type CreateLoop failed: route setup: EOF
[2019-05-16T12:24:21+03:00] INFO [trmanager]: Transport f422406e-2631-5659-98fe-1c277574063b failed with error: EOF. Re-dialing...
[2019-05-16T12:24:21+03:00] INFO [trmanager]: Closing transport manager
[2019-05-16T12:24:21+03:00] INFO [trmanager]: Transport f422406e-2631-5659-98fe-1c277574063b closed
[2019-05-16T12:24:21+03:00] INFO [trmanager]: Transport f422406e-2631-5659-98fe-1c277574063b closed
.....
[2019-05-16T12:24:21+03:00] INFO [trmanager]: Transport f422406e-2631-5659-98fe-1c277574063b closed
[2019-05-16T12:24:21+03:00] INFO [trmanager]: Transport f422406e-2631-5659-98fe-1c277574063b closed
[2019-05-16T12:24:21+03:00] INFO [trmanager]: Transport f422406e-2631-5659-98fe-1c277574063b closed
[2019-05-16T12:24:21+03:00] INFO [messenger]: Opened new link with the server 035915c609f71d0c7df27df85ec698ceca0cb262590a54f732e3bbd0cc68d89282
[2019-05-16T12:24:21+03:00] INFO [messenger]: Updating discovery entry
[...]

PR #379, opened 2019-05-28
Feature/sequence number in payload

### LINKED ISSUE #378: Append sequence in encryption payload
Instead of guessing the nonce append it to the payload.

ivcosla:
Closes #378

### DIFF
--- a/internal/noise/noise.go
+++ b/internal/noise/noise.go
@@ -4,10 +4,14 @@ import (
 	"crypto/rand"
 	"encoding/binary"
 
+	"github.com/skycoin/skycoin/src/util/logging"
+
 	"github.com/flynn/noise"
 
 	"github.com/skycoin/skywire/pkg/cipher"
 )
+
+var logger = logging.MustGetLogger("noise")
 
 // Config hold noise parameters.
 type Config struct {
@@ -29,7 +33,9 @@ type Noise struct {
 	enc     *noise.CipherState
 	dec     *noise.CipherState
 
-	seq uint32 // sequence number, used as nonce for both encrypting and decrypting
+	seq             uint32 // sequence number, used as nonce for both encrypting and decrypting
+	previousSeq     uint32 // sequence number last decrypted, check in order to avoid reply attacks
+	highestPrevious uint32 // highest sequence number received from the other end
 	//encN uint32 // counter to inform encrypting CipherState to re-key
 	//decN uint32 // counter to inform decrypting CipherState to re-key
 }
@@ -129,6 +135,16 @@ func (ns *Noise) EncryptUnsafe(plaintext []byte) []byte {
 // be used with external lock.
 func (ns *Noise) DecryptUnsafe(ciphertext []byte) ([]byte, error) {
 	seq := binary.BigEndian.Uint32(ciphertext[:4])
+	if seq <= ns.previousSeq {
+		logger.Warnf("current seq: %s is not higher than previous one: %s. "+
+			"Highest sequence number received so far is: %s", ns.seq, ns.previousSeq, ns.highestPrevious)
+	} else {
+		if ns.previousSeq > ns.highestPrevious {
+			ns.highestPrevious = seq
+		}
+		ns.previousSeq = seq
+	}
+
 	return ns.dec.Cipher().Decrypt(nil, uint64(seq), nil, ciphertext[4:])
 }
 
--- a/internal/noise/noise.go
+++ b/internal/noise/noise.go
@@ -2,14 +2,13 @@ package noise
 
 import (
 	"crypto/rand"
+	"encoding/binary"
+	"fmt"
 
 	"github.com/flynn/noise"
 
 	"github.com/skycoin/skywire/pkg/cipher"
 )
-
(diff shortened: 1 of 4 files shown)


pkg/messaging/channel_test.go
```diff
@@ -85,15 +85,15 @@ func TestChannelWrite(t *testing.T) {
...
+	buf := make([]byte, 29)
 	go out.Read(buf) // nolint
 	n, err := c.Write([]byte("foo"))
 	require.NoError(t, err)
 	assert.Equal(t, 3, n)
 
 	assert.Equal(t, FrameTypeSend, FrameType(buf[2]))
 	assert.Equal(t, byte(10), buf[3])
-	require.Equal(t, uint16(19), binary.BigEndian.Uint16(buf[4:]))
+	//require.Equal(t, uint16(19), binary.BigEndian.Uint16(buf[4:]))
```
ivcosla: Not sure what this was testing, it was failing now.

internal/noise/noise.go
```diff
@@ -31,8 +29,9 @@ type Noise struct {
 	enc     *noise.CipherState
 	dec     *noise.CipherState
 
-	encN uint32 // counter to inform encrypting CipherState to re-key
-	decN uint32 // counter to inform decrypting CipherState to re-key
+	seq uint32 // sequence number, used as nonce for both encrypting and decrypting
```
ivcosla: Note that now there is no rekey (which was leading to chacha errors). Seq value is also not being set to zero before it overflows.

ivcosla:
Tested using ssh integration environment. After running integration-startup I was able to run ssh without a single bug, however, ssh-integration-test failed to start ssh client correctly.

No data races were spotted during the test.

internal/noise/noise.go
evanlinjin: @ivcosla correct me if I'm wrong, shouldn't we have two `seq` numbers; one for encrypting `CipherState` and one for decrypting `CipherState`?

ivcosla: I didn't saw the need for that. We are taking the seq number from the payload when decrypting, so how would it be used?. We can keep track of how many decryptions we have performed with it, but it won't be used for encryption right?. If you have some suggestion regarding this, go for it!.

ayuryshev (approved):
It's obviously improves life of ssh-sessions.
Actually I tried to break it and was unsuccessful in this attempt.

That's great!

internal/noise/noise.go
evanlinjin: Ahhh I see. However this implementation way lead to security issues. We need to test for non-duplicate nonces before decrypting.

evanlinjin: As mentioned on Discord: We need to keep track of decryption nonces, and always ensure the next nonce > current nonce.

ivcosla:
```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9bafe8]

goroutine 97 [running]:
github.com/flynn/noise.(*CipherState).Cipher(...)
        /home/ivan/Escritorio/skycoin/pkg/mod/github.com/flynn/noise@v0.0.0-20180327030543-2492fe189ae6/state.go:61
github.com/skycoin/skywire/internal/noise.(*Noise).EncryptUnsafe(0xc000140750, 0xc0002a8380, 0x208, 0x30a, 0xc0002a2120, 0xc0002978f0, 0x589e93)
        /home/ivan/skycoin/skywire/internal/noise/noise.go:129 +0x1b8
github.com/skycoin/skywire/pkg/messaging.(*msgChannel).Write(0xc000106630, 0xc0002a8380, 0x208, 0x30a, 0x0, 0x0, 0x99c314f9b87fe5f2)
        /home/ivan/skycoin/skywire/pkg/messaging/channel.go:131 +0x310
encoding/json.(*Encoder).Encode(0xc000297ef0, 0xbdfdc0, 0xc000097860, 0x7fc4883bf320, 0xc000106630)
        /usr/lib/go/src/encoding/json/stream.go:225 +0x241
github.com/skycoin/skywire/pkg/transport.settlementInitiatorHandshake.func1(0xc00019ad90, 0xd59ea0, 0xc000106630, 0xc0003b41e0, 0x0, 0xc00006fce0)
        /home/ivan/skycoin/skywire/pkg/transport/handshake.go:46 +0x8d2
github.com/skycoin/skywire/pkg/transport.settlementHandshake.Do.func1(0xc0003deb00, 0xc00019ad90, 0xd59ea0, 0xc000106630, 0xc000010890, 0xc000139760, 0xc0003dc720)
        /home/ivan/skycoin/skywire/pkg/transport/handshake.go:18 +0x56
created by github.com/skycoin/skywire/pkg/transport.settlementHandshake.Do
        /home/ivan/skycoin/skywire/pkg/transport/handshake.go:17 +0x108
```

I ran into this once when closing node B. Not sure what can be nil by then on line 129.

ivcosla:
The referenced line in previous comment is actually this one: 
https://github.com/skycoin/skywire/pull/379/files#diff-9fba0b720ed6c60cb8dd622f7dc3edecR125

jdknives:
This also closes #355 ?

PR #380, opened 2019-05-29
Fix alex's PR race condition.

evanlinjin:


pkg/messaging/channel.go
```diff
@@ -237,7 +237,7 @@ func (mCh *msgChannel) readEncrypted(ctx context.Context, p []byte) (n int, err
 	}
 
 	if len(data) > len(p) {
-		if _, err := mCh.buf.Write(data[len(p):]); err != nil {
+		if _, err := mCh.buf.Write(data[len(p):]); err != nil { // TODO: data race.
```
ivcosla, line 240: Is there a data race now (new one or previously existing one)?

evanlinjin: I mean, after Alex's fix, the first message fails to deliver (due to some data race).

pkg/router/router.go
```diff
@@ -78,44 +78,32 @@ func New(config *Config) *Router {
...
-			}(tr)
-		}
-	}()
 
 	go func() {
-		for tr := range r.tm.DialedTrChan {
-			if r.IsSetupTransport(tr) {
+		for tp := range r.tm.TrChan {
+			var (
+				isAccepted = tp.Accepted
```
nkryuchkov: Is there any reason why an explicit variable declaration is used here? As for me, a `:=`-assignment would be more compact.

pkg/router/router_test.go
```diff
@@ -487,8 +487,14 @@ func TestRouterSetupLoop(t *testing.T) {
 	r := New(conf)
 	errCh := make(chan error)
 	go func() {
-		acceptCh, _ := m2.Observe()
-		tr := <-acceptCh
+		// acceptCh, _ := m2.Observe()
```
nkryuchkov: I'd consider deleting unused code instead of commenting it out (as well as other commented out code snippets)

pkg/router/router_test.go
```diff
@@ -487,8 +487,14 @@ func TestRouterSetupLoop(t *testing.T) {
...
+		// tr := <-acceptCh
+		var tr *transport.ManagedTransport
```
nkryuchkov, line 490: I think we could get rid of this line rewriting the line below to `for tr := range m2.TrChan {` (as well as in other tests)

pkg/setup/node.go
```diff
@@ -91,22 +91,38 @@ func (sn *Node) Serve(ctx context.Context) error {
...
 
 	go func() {
-		for tr := range sn.tm.AcceptedTrChan {
-			go func(t transport.Transport) {
-				for {
-					if err := sn.serveTransport(t); err != nil {
-						sn.Logger.Warnf("Failed to serve Transport: %s", err)
-						return
+		for tr := range sn.tm.TrChan {
+
```
nkryuchkov: I'd consider removing this empty line

pkg/messaging/channel.go
evanlinjin: So this is fixed.

nkryuchkov (approved):
Looks good to me. We could consider making changes on the lines where I left comments

pkg/router/router.go
evanlinjin: Sorry it's my bad habit. I like it when variable declarations line up haha. I'll change this.

nkryuchkov: @evanlinjin No problem, it's just my opinion. Of course, it's considerable.

pkg/router/router_test.go
evanlinjin: `tr` is used below. Hence, this is still needed.

nkryuchkov: @evanlinjin Yes, you're right. I missed the usages below. Sorry

ivcosla (approved):

ayuryshev (approved):
It's indeed cures PR #376 
That's great!

PR #382, opened 2019-05-30
[WIP] 'dmsg' implementation.

evanlinjin:


pkg/messaging/client.go
```diff
@@ -289,60 +289,60 @@ func (c *Client) onData(l *Link, frameType FrameType, body []byte) error {
...
-		clientLink.chans.remove(channelID)
-		_, sendErr = l.SendChannelClosed(channel.ID())
-		c.Logger.Debugf("Closed channel ID %d", channelID)
+		clientLink.chans.remove(localID)
+		_, sendErr = l.SendChannelClosed(localID)
+		c.Logger.Debugf("Closed channel ID %d", localID)
 	case FrameTypeChannelOpened:
-		channel.SetID(body[1])
-		if err := channel.ProcessMessage(body[2:]); err != nil {
+		remote.SetID(body[1])
```
ayuryshev: Can't we rid off `SetID`?

jdknives:
@evanlinjin should be re-named Skycoin DMS (distributed messaging system)

ayuryshev (commented):
I would consider not to make changes in `pkg/messaging` but make an alternative implementation of `transport.Transport`
Possibly as fork of current `pkg/messaging`

pkg/messaging/client.go
evanlinjin: changes to `/pkg/messaging` are accidental.

pkg/dms/channel.go
```diff
@@ -0,0 +1,166 @@
...
+func (c *Channel) close() bool {
+	closed := false
+	c.doneOnce.Do(func() {
+		close(c.doneCh)
+		closed = true
+	})
+	return closed
+}
+
+func (c *Channel) Handshake(ctx context.Context) error {
```
ivcosla: Great!

pkg/dms/channel.go
```diff
@@ -0,0 +1,166 @@
...
+	// if channel ID is even, client is initiator.
+	if init := isEven(c.id); init {
```
ayuryshev:
```golang
if (c.id%2 ==0) {
}
```
?
IMO separate function is not needed for this check

pkg/dms/channel.go
```diff
@@ -0,0 +1,166 @@
...
+	"io"
+	"net"
+	"sync"
+
+	"github.com/skycoin/skywire/pkg/cipher"
+	"github.com/skycoin/skywire/pkg/transport"
+)
+
+// Channel is a channel from a client's perspective.
+type Channel struct {
```
ayuryshev: IMO it's better to avoid such generic words as "channel" in names.
Does this  `Channel` implements `Transport`?
If so: let's name it `DMSTransport` or similar.

pkg/dms/channel.go
```diff
@@ -0,0 +1,166 @@
...
+	case <-c.doneCh:
+		return errors.New("closed")
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (c *Channel) close() bool {
+	closed := false
+	c.doneOnce.Do(func() {
```
ayuryshev: Good!

pkg/dms/client.go
```diff
@@ -0,0 +1,385 @@
...
+	// ErrChannelClosed indicates that underlying channel is being closed and writes are prohibited.
+	ErrChannelClosed = errors.New("channel closed")
+	// ErrDeadlineExceeded indicates that read/write operation failed due to timeout.
+	ErrDeadlineExceeded = errors.New("deadline exceeded")
+	// ErrClientClosed indicates that client is closed and not accepting new connections.
+	ErrClientClosed = errors.New("client closed")
+)
+
+// Link from a client's perspective.
+type Link struct {
```
ayuryshev: We have misleading naming here:
- in networks "link" is L2 protocol, one can expect WiFi, Ethernet and such here
- but this `Link` embeds `net.Conn` which is 2 layers above, such as  is TCP, UDP 

`DMSConn`?

pkg/dms/client.go
```diff
@@ -0,0 +1,385 @@
...
+			_ = ch.Close()
+		}
+	}
+	err := l.Conn.Close()
+	l.mx.Unlock()
+	l.wg.Wait()
+	return err
+}
+
+type Client struct {
```
ayuryshev: `Client` implements `transport.Factory`?

cmd/messaging-server/commands/root.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"bufio"
+	"encoding/json"
+	"io"
+	"log"
+	"log/syslog"
+	"net/http"
+	"os"
+
+	"github.com/skycoin/skywire/pkg/dms"
```
nkryuchkov: What do you think about moving this line between lines 19 and 20 to make all `github.com/skycoin/skywire` imports grouped?

internal/ioutil/len_read_writer.go
```diff
@@ -4,34 +4,38 @@ import (
...
+	return &LenReadWriter{ReadWriter: rw}
 }
 
 // ReadPacket returns single received len prepended packet.
-func (rw *LenReadWriter) ReadPacket() (data []byte, err error) {
-	var size uint16
-	if err = binary.Read(rw.ReadWriter, binary.BigEndian, &size); err != nil {
-		return
+func (rw *LenReadWriter) ReadPacket() ([]byte, error) {
+	h := make([]byte, 2)
```
nkryuchkov, line 26: What do you think about extracting `2` to a constant? I think it might be more readable. We could also remove other occurrences of `2`s.

internal/noise/net.go
```diff
@@ -205,23 +205,25 @@ func WrapListener(lis net.Listener, pk cipher.PubKey, sk cipher.SecKey, init boo
...
-		LocalSK:   ml.sk,
-		Initiator: ml.init,
-	})
-	if err != nil {
-		return nil, err
-	}
-	rw := NewReadWriter(conn, ns)
-	if err := rw.Handshake(time.Second * 10); err != nil {
-		return nil, err
+	for {
```
nkryuchkov, line 208: Is an infinite loop possible here if either `New` or `rw.Handshake` would always return an error? Would adding some debug level logs with returned errors help?

pkg/dms/channel.go
evanlinjin: It's not needed, but I like it that way.

pkg/dms/client.go
evanlinjin: Yes.

pkg/dms/channel.go
```diff
@@ -0,0 +1,162 @@
...
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"net"
+	"sync"
+
+	"github.com/skycoin/skywire/internal/ioutil"
+
```
nkryuchkov: I'd consider deleting this empty line

pkg/dms/channel.go
nkryuchkov: DMSTransport starts with the package name, so if we rename it to something that contains the word `Transport`, I'd name it just `Transport`.

pkg/dms/channel.go
nkryuchkov: I think the separate function is fine because it doesn't hurt the readability. 

Moreover, we can change its implementation later.
E.g. we can rewrite it to
```
if (c.id&1 == 0) {
}
``` 
later to increase the performance a bit, because the Go compiler optimization layer is not that fine and may compile the modulo operation to division instructions, not logic ones.

pkg/dms/channel.go
```diff
@@ -0,0 +1,162 @@
...
+			return errors.New("invalid remote response")
+		}
+	case <-c.doneCh:
+		return errors.New("closed")
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (c *Channel) close() bool {
```
nkryuchkov: I'd consider making the returned boolean variable named: `func (c *Channel) close() (closed bool) {`. This would let understand what the function returns reading just a signature.

pkg/dms/frame.go
```diff
@@ -0,0 +1,78 @@
...
+
+const (
+	TpType = "dms"
+
+	hsTimeout  = time.Second * 10
+	readBufLen = 10
+	headerLen  = 5 // fType(1 byte), chID(2 byte), payLen(2 byte)
+)
+
+func isEven(chID uint16) bool { return chID%2 == 0 }
```
nkryuchkov: Could we rename it to `isInitiator`?

pkg/dms/frame.go
```diff
@@ -0,0 +1,78 @@
...
+	CloseType   = FrameType(3)
+	FwdType     = FrameType(10)
+)
+
+type Frame []byte
+
+func MakeFrame(ft FrameType, chID uint16, pay []byte) Frame {
+	f := make(Frame, headerLen+len(pay))
+	f[0] = byte(ft)
+	binary.BigEndian.PutUint16(f[1:3], chID)
```
nkryuchkov: What do you think if we extract the constants like `channelIDOffset` `channelIDLength`?

pkg/dms/server.go
```diff
@@ -0,0 +1,343 @@
...
+	l.mx.Lock()
+	l.relations[id] = r
+	l.mx.Unlock()
+}
+
+func (l *ServerLink) getRelation(id uint16) (*Relation, bool) {
+	l.mx.RLock()
+	r := l.relations[id]
+	l.mx.RUnlock()
+	return r, r != nil
```
nkryuchkov: Would it be more readable if the caller checks if `r != nil` instead of checking the second returned value?

pkg/dms/server.go
```diff
@@ -0,0 +1,343 @@
...
+		s.wg.Add(1)
+		go func() {
+			defer s.wg.Done()
+			_ = link.Serve(ctx, s.in) // TODO: log error.
+		}()
+	}
+}
+
+func (s *Server) serve() {
+
```
nkryuchkov: Could we remove the empty line here?

pkg/node/node.go
```diff
@@ -15,11 +15,12 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/skycoin/skywire/pkg/dms"
```
nkryuchkov: Could we group it with the lines below?

pkg/setup/node.go
```diff
@@ -8,12 +8,13 @@ import (
 	"log"
 	"time"
 
+	"github.com/skycoin/skywire/pkg/dms"
```
nkryuchkov: I'd consider grouping the `github.com/skycoin/skywire/` imports

pkg/setup/node_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/skycoin/skywire/pkg/dms"
+
```
nkryuchkov, line 12: Could we remove the empty line here?

pkg/transport/managed_transport.go
```diff
@@ -88,11 +91,13 @@ func (tr *ManagedTransport) Write(p []byte) (n int, err error) {
 func (tr *ManagedTransport) Close() error {
 	tr.mu.RLock()
 	err := tr.Transport.Close()
+	tr.isClosing = true
 	tr.mu.RUnlock()
 
 	select {
 	case <-tr.doneChan:
 	default:
+
```
nkryuchkov: Could we remove the empty line here?

internal/noise/net.go
evanlinjin: That’s the desired functionality. Logs may help.

pkg/dms/server.go
evanlinjin: From my understanding, this is common Golang convention to use a second output as bool or error.

pkg/dms/frame.go
evanlinjin: I think they should be constants if they are to be used a large number of times. I think comments will be more suitable in this scenario.

internal/ioutil/len_read_writer.go
evanlinjin: I think this is a matter of taste. Personally I prefer if it’s left like this. 2 bytes is encoded size for uint16 and not some difficult value.

internal/ioutil/buf_read.go
```diff
@@ -0,0 +1,16 @@
+package ioutil
+
+import (
+	"bytes"
+)
+
+// BufRead is designed to help writing 'io.Reader' implementations.
+// It reads from 'data' into 'p'. If 'p' is short, write to 'buf'.
+// Note that one should check if 'buf' has data and read from that first before calling this function.
+func BufRead(buf *bytes.Buffer, data, p []byte) (int, error) {
```
ivcosla, line 10: Good idea!

pkg/setup/node_test.go
ayuryshev: IMO it would better to leave for `make format`

pkg/dms/client.go
```diff
@@ -0,0 +1,386 @@
...
+	c.mx.RUnlock()
+	ok := tp != nil && !tp.IsDone()
+	return tp, ok
+}
+
+// local:  local client pk  (also responding client).
+// remote: remote client pk (also initiating client).
+func checkRequest(local cipher.PubKey, id uint16, p []byte) (remote cipher.PubKey, ok bool) {
+	// server-initiated channels should have odd channel ID
+	if isEven(id) {
```
ivcosla: then that would be better renamed as `isServerInitiated`. Just a suggestion, the comment clarifies.

pkg/node/node.go
```diff
@@ -197,15 +198,12 @@ func NewNode(config *Config) (*Node, error) {
 // Start spawns auto-started Apps, starts router and RPC interfaces .
 func (node *Node) Start() error {
 	ctx := context.Background()
-	err := node.messenger.ConnectToInitialServers(ctx, node.config.Messaging.ServerCount)
+	err := node.messenger.InitiateServers(ctx, node.config.Messaging.ServerCount)
```
ivcosla: Maybe `InitiateServerConnections` describes functionality better.

PR #384, opened 2019-05-30
[M2] [WIP] renamed and moved skywire ssh

ivcosla:
WIP until integration tests on skywire-services integrate this changes.


evanlinjin:
`SSH` in capitals is rather odd in my opinion. Is this a specific request from Brandon? Why aren't we naming it `dmsg-ssh` or `skyssh`?

gz-c:
no mixed cases

jdknives:
@gz-c @evanlinjin `skyssh` would be fine. We should not name it `dmsg-ssh` since conceptually it is not related according to Synth.

ayuryshev (approved):

jdknives:
This is ready for merging?

PR #393, opened 2019-06-06
Integration scripts for skywire with public services

ayuryshev:
Ported generic environment from skywire-services

DONE:

- [x] skychat
- [x] socksproxy
- [x] SSH
- [x] Makefile goals
- [x] documentation

Diagnostic scripts:

- integration/check-services.sh - checks availability of messaging-discovery, transport-discovery and route-finder services
- integration/check-route-finder.sh  - checks routes $PK_A-$PK_B-$PK_C in route-finder

Closes #387

PR #399, opened 2019-06-09
Feature: embedded profiling of skywire-node

ayuryshev:
# Feature: embedded profiler in skywire-node

### 1. Implemented ability to collect *.pprof files on exit of skywire-node

To create profiling data add flag `--pprof/-p` for skywire-node.

Possible values:
- cpu
- mem
- mutex
- block

pprof-files will be generated on exit in ./logs/[--tag]/[mode].pprof

To generate a report use `go tool pprof`

E.g. 

```bash
$ ./bin/skywire-node ./integration/generic/nodeA.json --tag NodeA --pprof
cpu
# Run tests
# then Ctrl-C
$ go tool pprof --pdf ~/bin/skywire-node ./logs/nodeA/cpu.pprof
```

### 2. Interactive profiling

Run skywire-node with `--pprof http --pport 6060`
(--pport has default value 6060 and could be ommitted when only one node
profiled)

E.g.
```bash 
$ skywire-node ./integration/generic/nodeA.json --tag NodeA --pprof http
$ go tool pprof http://localhost:6060/debug/pprof/
```

Read: 

- https://golang.org/pkg/net/http/pprof/
- https://blog.golang.org/profiling-go-programs 
- https://godoc.org/github.com/pkg/profile

### 1. Environment variable $BUILD_OPTS with default value: "-race"

Motivation:

Need to switch on/off `-race` flag for various tests and various profiles.

Now it's possible:

```bash
$ export BUILD_OPTS=
$ make build  # or `make integration-build`
# note that binaries built without -race flag 
```

2. Environment variable $TEST_OPTS with default value: "-race -tags no_ci -cover -timeout=5m"

Motivation:

- `-timeout=5m` - is a good, safe timeout  for travis-ci 
- for developer environments it's more comfortable to use 60s (maximum
90s)

Now it's possible:

```bash 
$ export TEST_OPTS="-race -tags no_ci -timeout=60s"                                                                                                  5:41
$ make test
# note that timeout is not 5m now
```

3. Tidying up `cobra.Command.Run` in `cmd/skywire-node/command/root.go`

Code of `Run`-function was quite big even before and adding profiler option - made code even less readable.

After decomposition Run-function into 6 functions now it looks this way:

https://github.com/skycoin/skywire/blob/bf42797ed08a6046faf8b5357a01555a218fc03c/cmd/skywire-node/commands/root.go#L158
```golang
	Run: func(_ *cobra.Command, args []string) {
		cfg.args = args

		cfg.startProfiler().
			startLogger().
			readConfig().
			runNode().
			waitOsSignals().
			stopNode()
	},
```

Closes: #400


integration/InteractiveEnvironments.md
```diff
@@ -0,0 +1,274 @@
+# Interactive test environments
```
ivcosla: Very uselful file. Good job!

ayuryshev: It's because this PR is based on https://github.com/skycoin/skywire/pull/393

cmd/skywire-node/commands/root.go
```diff
@@ -16,92 +17,165 @@ import (
...
 
-var rootCmd = &cobra.Command{
-	Use:   "skywire-node [config-path]",
-	Short: "App Node for skywire",
-	Run: func(_ *cobra.Command, args []string) {
+var cfg *runCfg
 
-		logger := logging.MustGetLogger(tag)
+func (cfg *runCfg) startProfiler() *runCfg {
+	var option func(*profile.Profile)
```
Darkren, line 49: Wouldn't it be better if we could write several profiles at once? For example, `cpu` and `mem`. There are a situations when sometimes it hangs, sometimes it doesn't, it is not that comfortable to get profiles from several runs, they may not show the whole picture

Darkren (changes requested):

PR #402, opened 2019-06-10
Implementation of dmsg.

evanlinjin:
This is a renaming of #382


pkg/dmsg/README.md
```diff
@@ -0,0 +1,276 @@
...
+| `0x3` | `CLOSE` | 1 byte that represents the reason for closing | 1 |
+| `0xa` | `FWD` | uint16 sequence + transport payload | >2 |
+| `0xb` | `ACK` | uint16 sequence | 2 |
+
+## Transports
+
+Transports are represented by transport IDs and facilitate duplex communication between two `dmsg.Client`s which are connected to a common `dmsg.Server`.
+
+Transport IDs are assigned in such a manner:
+- A `dmsg.Client` manages the assignment of positive transport IDs between itself and each connected `dmsg.Server`. The set of transport IDs will be unique between itself and each `dmsg.Server`.
```
ivcosla: Here we talk about positive/negative transports IDs, while below we do so about even/odd tranport IDs

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,352 @@
...
+// Errors related to REQUEST frames.
+var (
+	ErrRequestRejected    = errors.New("failed to create transport: request rejected")
+	ErrRequestCheckFailed = errors.New("failed to create transport: request check failed")
+	ErrAcceptCheckFailed  = errors.New("failed to create transport: accept check failed")
+)
+
+// Transport represents a connection from dmsg.Client to remote dmsg.Client (via dmsg.Server intermediary).
+// It implements transport.Transport
+type Transport struct {
```
ivcosla, line 27: Transport has a lot of fields and what they should be used for can be therefore confusing, I would suggest to cluster the ones that are related together in new and small objects if possible. 

However take this as a comment, I would leave it for later since is not prioritary, but definitely do it in the future.

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,352 @@
...
+
+		default:
+			_ = tp.Close() //nolint:errcheck
+			return ErrAcceptCheckFailed
+		}
+	}
+}
+
+// Serve handles received frames.
+func (tp *Transport) Serve() {
```
ivcosla, line 228: This function is also something to simplify in the future

pkg/dmsg/client_test.go
```diff
@@ -0,0 +1,288 @@
...
+		conn4.setNextInitID(randID(false))
+
+		ch1 := make(chan *Transport, acceptChSize)
+		ch2 := make(chan *Transport, acceptChSize)
+		ch3 := make(chan *Transport, acceptChSize)
+		ch4 := make(chan *Transport, acceptChSize)
+
+		ctx := context.TODO()
+
+		go func() {
```
ivcosla, line 118: A lot of patterns are repeated in these two functions, like the conn.Serve go routines, conn initialization, etc. I would suggest to extract some of them as functions to improve readability.

For example:

```go
conn1, conn2, conn3, conn4 := setupTransports()
```

ivcosla (changes requested):
Nice work overall, just a few things to check.

pkg/dmsg/README.md
evanlinjin: Sorted.

pkg/dmsg/transport.go
evanlinjin: @ivcosla I would have to disagree on this one. The logic is simple, we are handling incoming frames, and handling them based on it's type. I would say that maybe some comments on the logic would make things easier though.

ivcosla: I agree is not difficult, just very long, well, up to you.

pkg/dmsg/transport.go
evanlinjin: @ivcosla I've added comments to the fields as a compromise. As the code-base for `dmsg.Transport` is relatively small, I feel that splitting it up into multiple structures will not make the code easier to read.

ivcosla: That might be enough, good job.

Makefile
```diff
@@ -105,9 +105,11 @@ bin: ## Build `skywire-node`, `skywire-cli`, `manager-node`, `SSH-cli`
 	${OPTS} go build -race -o ./skywire-node ./cmd/skywire-node 
 	${OPTS} go build -race -o ./skywire-cli  ./cmd/skywire-cli 
 	${OPTS} go build -race -o ./setup-node ./cmd/setup-node
+	${OPTS} go build -race -o ./messaging-server ./cmd/messaging-server
```
ayuryshev, line 109: What is the reason to move messaging-server from skywire-services?
Is it temporary measure?

internal/ioutil/ack_waiter_test.go
```diff
@@ -0,0 +1,45 @@
+package ioutil
```
ayuryshev: Consider make this tests blackbox: only public API is tested here

```golang
package ioutil_test

import (
	"context"
	"sync"
	"testing"

	"github.com/skycoin/skywire/internal/ioutil"
)
func TestUint16AckWaiter_Wait(t *testing.T) {

	// Ensure that no race conditions occurs when
	// each concurrent call to 'Uint16AckWaiter.Wait()' is met with
	// multiple concurrent calls to 'Uint16AckWaiter.Done()' with the same seq.
	t.Run("ensure_no_race_conditions", func(*testing.T) {
		w := new(ioutil.Uint16AckWaiter)
		defer w.StopAll()

		seqChan := make(chan ioutil.Uint16Seq)
		defer close(seqChan)

		wg := new(sync.WaitGroup)

		for i := 0; i < 64; i++ {
			wg.Add(1)
			go func() {
				defer wg.Done()
				_ = w.Wait(context.TODO(), func(seq ioutil.Uint16Seq) error { //nolint:errcheck,unparam
					seqChan <- seq
					return nil
				})
			}()

			seq := <-seqChan
			for j := 0; j <= i; j++ {
				wg.Add(1)
				go func() {
					defer wg.Done()
					w.Done(seq)
				}()
			}
		}

		wg.Wait()
	})
}

```

internal/ioutil/atomic_bool.go
```diff
@@ -0,0 +1,23 @@
+package ioutil
+
+import "sync/atomic"
+
+// AtomicBool implements a thread-safe boolean value.
+type AtomicBool struct {
```
ayuryshev, line 6: I can't find usage of this type.
I consider to drop it.

internal/ioutil/buf_read.go
```diff
@@ -0,0 +1,16 @@
+package ioutil
+
+import (
+	"bytes"
+)
+
+// BufRead is designed to help writing 'io.Reader' implementations.
+// It reads from 'data' into 'p'. If 'p' is short, write to 'buf'.
+// Note that one should check if 'buf' has data and read from that first before calling this function.
+func BufRead(buf *bytes.Buffer, data, p []byte) (int, error) {
```
ayuryshev, line 10: I see only one usage of this function.

Does it deserves a separate file?

internal/noise/net.go
```diff
@@ -205,23 +205,27 @@ func WrapListener(lis net.Listener, pk cipher.PubKey, sk cipher.SecKey, init boo
...
+		ns, err := New(ml.pattern, Config{
+			LocalPK:   ml.pk,
+			LocalSK:   ml.sk,
+			Initiator: ml.init,
+		})
+		if err != nil {
+			return nil, err
+		}
+		rw := NewReadWriter(conn, ns)
+		if err := rw.Handshake(time.Second * 10); err != nil {
```
ayuryshev: `time.Second * 10` is meant to be configurable?

internal/noise/noise.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/skycoin/skywire/pkg/cipher"
 )
 
-var logger = logging.MustGetLogger("noise")
+var log = logging.MustGetLogger("noise")
```
ayuryshev, line 14: `var log` is used inside "github.com/skycoin/skycoin/src/util/logging" for `MasterLogger`.

Althought it's syntactically correct I would consider to use `logger` or `noiseLogger` here.

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+>**TODO:**
+>
+>- `ACK` frames should include the first 4 bytes of the rolling hash of incoming payloads, enforcing reliability of data. Transports should therefore keep track of incoming/outgoing rolling hashes.
+>- Transports should also be noise-encrypted. `REQUEST` and `ACCEPT` frames should include noise handshake messages (`KK` handshake pattern), and the `FWD` and `ACK` payloads are to be encrypted.
+> - Transports should implement read/write deadlines and local/remote addresses (like `net.Conn`).
+> - `dmsg.Server` should check incoming frames to disallow excessive sending of `CLOSE`, `ACCEPT` and `REQUEST` frames.
+
+## Terminology
+
+- **entity -** A service of `dmsg` (typically being part of an executable running on a machine).
```
ayuryshev, line 12: Explanation starts with "A service"
So why not use "service" instead of vague "entity"

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+- **entity type -** The type of entity. `dmsg` has three entity types; `dmsg.Server`, `dmsg.Client`, `dmsg.Discovery`.
```
ayuryshev, line 13: **service type** ? 
"has three service types:" ?

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+- **entry -** A data structure that describes an entity and is stored in `dmsg.Discovery` for entities to access.
```
ayuryshev, line 14: It clashes with Entry-types from transport.
And here meaning "describes dmsg service" - right?

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+- **frame -** The data unit of the `dmsg` system.
+- **frame type -** The type of `dmsg` frame. There are four frame types; `REQUEST`, `ACCEPT`, `CLOSE`, `FWD`, `ACK`.
+- **connection -** The direct line of duplex communication between a `dmsg.Server` and `dmsg.Client`.
+- **transport -** A line of communication between two `dmsg.Client`s that is proxied via a `dmsg.Server`.
+- **transport ID -** A uint16 value that identifies a transport.
```
ayuryshev, line 19: Is it implementation detail or part of public API?

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+> - `dmsg.Server` should check incoming frames to disallow excessive sending of `CLOSE`, `ACCEPT` and `REQUEST` frames.
+
+## Terminology
+
+- **entity -** A service of `dmsg` (typically being part of an executable running on a machine).
+- **entity type -** The type of entity. `dmsg` has three entity types; `dmsg.Server`, `dmsg.Client`, `dmsg.Discovery`.
+- **entry -** A data structure that describes an entity and is stored in `dmsg.Discovery` for entities to access.
+- **frame -** The data unit of the `dmsg` system.
+- **frame type -** The type of `dmsg` frame. There are four frame types; `REQUEST`, `ACCEPT`, `CLOSE`, `FWD`, `ACK`.
+- **connection -** The direct line of duplex communication between a `dmsg.Server` and `dmsg.Client`.
```
ayuryshev, line 17: I never saw "line of communication" used for networks.
Googled - it looks like military/business terminology.

Why not just change to "network connection"?

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+- **transport -** A line of communication between two `dmsg.Client`s that is proxied via a `dmsg.Server`.
+- **transport ID -** A uint16 value that identifies a transport.
+
+## Entities
+
+The `dmsg` system is made up of three entity types:
+- `dmsg.Discovery` is a RESTful API that allows `dmsg.Client`s to find remote `dmg.Client`s and `dmsg.Server`s.
+- `dmsg.Server` proxies frames between clients.
```
ayuryshev, line 25: "proxies frames" - "frames" here "packets"?

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+- The `Payload` have a size that is obtainable via `PayloadSize`.
+
+The following is a summary of the frame types:
+
+| FrameType | Name | Payload Contents | Payload Size |
+| --- | --- | --- | --- |
+| `0x1` | `REQUEST` | initiating client's public key + responding client's public key | 66 |
+| `0x2` | `ACCEPT` | initiating client's public key + responding client's public key | 66 |
+| `0x3` | `CLOSE` | 1 byte that represents the reason for closing | 1 |
+| `0xa` | `FWD` | uint16 sequence + transport payload | >2 |
```
ayuryshev, line 73: Why `0xa` here?

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+Transport IDs are assigned in such a manner:
+- A `dmsg.Client` manages the assignment of even transport IDs between itself and each connected `dmsg.Server`. The set of transport IDs will be unique between itself and each `dmsg.Server`.
+- A `dmsg.Server` manages the assignment of odd transport IDs between itself and each connected `dmsg.Client`. The set of transport IDs will be unique between itself and each `dmsg.Client`.
+
+For a given transport:
+- Between the initiating client and the common server - the transport ID is always a even value.
+- Between the responding client and the common server - the transport ID is always a odd value.
+
+Hence, a transport in it's entirety, is represented by 2 transport IDs.
```
ayuryshev, line 88: `transport.Transport` already has public `Edges()` which are 2 IDs.
And `dmgs.Transport` too.

Those `uint16` IDs:
- has only internal meaning for dmsg
- part of public API
?

pkg/dmsg/README.md
```diff
@@ -0,0 +1,277 @@
...
+3. The responding client receives the `REQUEST` frame and checks the contents. If valid, the responding client sends an `ACCEPT` frame (containing the same payload as the `REQUEST`) back to the common server. The common server changes the transport ID, and forwards the `ACCEPT` to the initiating client.
+
+On any step, if an error occurs, any entity can send a `CLOSE` frame.
+
+### Acknowledgement Logic
+
+Each `FWD` frame is to be met with an `ACK` frame in order to be considered delivered.
+
+- Each `FWD` payload has a 2-byte prefix (represented by a uint16 sequence). This sequence is unique per transport.
+- The destination of the transport, after receiving the `FWD` frame, responds with an `ACK` frame with the same sequence as the payload.
```
ayuryshev, line 103: ACK frame does not contains some checksum of FWD payload?
It's always successful?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+	net.Conn                // conn to dmsg server
+	local     cipher.PubKey // local client's pk
+	remoteSrv cipher.PubKey // dmsg server's public key
+
+	// nextInitID keeps track of unused tp_ids to assign a future locally-initiated tp.
+	// locally-initiated tps use an even tp_id between local and intermediary dms_server.
+	nextInitID uint16
+
+	// Transports: map of transports to remote dms_clients (key: tp_id, val: transport).
+	tps [math.MaxUint16 + 1]*Transport
```
ayuryshev, line 44: 1. `tps` has strong association with "transactions-per-second"
I would consider to rename it.

2. `tps [math.MaxUint16 + 1]*Transport` - what is the purpose of creating array of 65536 elements beforehand?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+		done:       make(chan struct{}),
+	}
+	cc.wg.Add(1)
+	return cc
+}
+
+// RemotePK returns the remote Server's PK that the ClientConn is connected to.
+func (c *ClientConn) RemotePK() cipher.PubKey { return c.remoteSrv }
+
+func (c *ClientConn) getNextInitID(ctx context.Context) (uint16, error) {
```
ayuryshev, line 69: Why not atomic AddUInt here?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+	c.mx.RLock()
+	tp := c.tps[id]
+	c.mx.RUnlock()
+	ok := tp != nil && !tp.IsClosed()
+	return tp, ok
+}
+
+func (c *ClientConn) setNextInitID(nextInitID uint16) {
+	c.mx.Lock()
+	c.nextInitID = nextInitID
```
ayuryshev, line 124: Again: I would consider atomics here

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+	c.mx.Unlock()
+}
+
+func (c *ClientConn) handleRequestFrame(accept chan<- *Transport, id uint16, p []byte) (cipher.PubKey, error) {
```
ayuryshev, line 142: `p []byte` here - payload?
I would consider to add several symbols here.

`data`?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+	}
+	if err := tp.ReadAccept(ctx); err != nil {
+		return nil, err
+	}
+	go tp.Serve()
+	return tp, nil
+}
+
+// Close closes the connection to dms_server.
+func (c *ClientConn) Close() error {
```
ayuryshev, line 258: Is it safe to repeat `c.Close()` multiple times?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+			if tp != nil {
+				go tp.Close() //nolint:errcheck
+			}
+		}
+		_ = c.Conn.Close() //nolint:errcheck
+		c.mx.Unlock()
+		c.wg.Wait()
+	})
+
+	if !closed {
```
ayuryshev: So the second c.Close() will return ErrClientClosed?
What is the meaning of this error here?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+		sk:     sk,
+		dc:     dc,
+		conns:  make(map[cipher.PubKey]*ClientConn),
+		accept: make(chan *Transport, acceptChSize),
+		done:   make(chan struct{}),
+	}
+}
+
+// SetLogger sets the dms_client's logger.
+func (c *Client) SetLogger(log *logging.Logger) {
```
ayuryshev: Why not assign logger only once - in NewClient?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+	c.log = log
+}
+
+func (c *Client) updateDiscEntry(ctx context.Context) error {
```
ayuryshev, line 324: This method is thread-unsafe by design?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+			return err
+		}
+		return c.dc.SetEntry(ctx, entry)
+	}
+	entry.Client.DelegatedServers = srvPKs
+	c.log.Infoln("updatingEntry:", entry)
+	return c.dc.UpdateEntry(ctx, c.sk, entry)
+}
+
+func (c *Client) setConn(ctx context.Context, l *ClientConn) {
```
ayuryshev: `l *ClientConn` - why `l` ?

pkg/dmsg/client.go
```diff
@@ -0,0 +1,525 @@
...
+		return nil, err
+	}
+	nc, err := noise.WrapConn(tcpConn, ns, hsTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	conn := NewClientConn(c.log, nc, c.pk, srvPK)
+	c.setConn(ctx, conn)
+	go func() {
```
ayuryshev, line 461: This gouroutine deserves to be named IMO.

```golang
func (c *Client) serve(ctx context.Context, conn *ClientConn, srvPK cipher.PubKey) {
	err := conn.Serve(ctx, c.accept)
	conn.log.WithError(err).WithField("remoteServer", srvPK).Warn("connected with server closed")
	c.delConn(ctx, srvPK)

	// reconnect logic.
retryServerConnect:
	select {
	case <-c.done:
	case <-ctx.Done():
	case <-time.After(time.Second * 3):
		conn.log.WithField("remoteServer", srvPK).Warn("Reconnecting")
		if _, err := c.findOrConnectToServer(ctx, srvPK); err != nil {
			conn.log.WithError(err).WithField("remoteServer", srvPK).Warn("ReconnectionFailed")
			goto retryServerConnect
		}
		conn.log.WithField("remoteServer", srvPK).Warn("ReconnectionSucceeded")
	}
}
```
?

pkg/dmsg/client_test.go
```diff
@@ -0,0 +1,288 @@
...
+		var pk1, pk2 cipher.PubKey
+		err := pk1.Set("024ec47420176680816e0406250e7156465e4531f5b26057c9f6297bb0303558c7")
+		assert.NoError(t, err)
+		err = pk2.Set("031b80cd5773143a39d940dc0710b93dcccc262a85108018a7a95ab9af734f8055")
+		assert.NoError(t, err)
+
+		conn1 := NewClientConn(logger, p1, pk1, pk2)
+		conn2 := NewClientConn(logger, p2, pk2, pk1)
+
+		conn2.setNextInitID(randID(false))
```
ayuryshev: Why this needed?

pkg/dmsg/client_test.go
```diff
@@ -0,0 +1,288 @@
...
+			errCh2 <- tr2.Close()
+		}()
+
+		err = <-errCh1
+		assert.NoError(t, err)
+
+		err = <-errCh2
+		assert.NoError(t, err)
+
+		go func() {
```
ayuryshev, line 223: Why not use just one `errCh` here?
And 
```golang
		for _, conn := range []*ClientConn{conn1, conn2, conn3, conn4} {
			errCh1 <- conn.Close()
		}
```

pkg/dmsg/client_test.go
```diff
@@ -0,0 +1,288 @@
...
+func isReadChannelOpen(ch chan Frame) bool {
+	select {
+	case _, ok := <-ch:
+		return ok
+	case <-time.After(chanReadThreshold):
+		return false
+	}
+}
+
+type invertedIDConn struct {
```
ayuryshev, line 279: What is the meaning of this type?

pkg/dmsg/client_test.go
```diff
@@ -0,0 +1,288 @@
...
+	net.Conn
+}
+
+func (c invertedIDConn) Write(b []byte) (n int, err error) {
+	frame := Frame(b)
+	newFrame := MakeFrame(frame.Type(), frame.TpID()^1, frame.Pay())
```
ayuryshev, line 286: InvertedIDConn will be even for odd and otherwise?

pkg/dmsg/frame.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/skycoin/skywire/internal/ioutil"
+
+	"github.com/skycoin/skywire/pkg/cipher"
+)
+
+const (
+	// Type returns the transport type string.
+	Type = "dmsg"
+
+	hsTimeout     = time.Second * 10
```
ayuryshev: `time.Second*10` - configuration value?

pkg/dmsg/frame.go
```diff
@@ -0,0 +1,160 @@
...
+	tpBufCap      = math.MaxUint16
+	tpBufFrameCap = math.MaxUint8
+	tpAckCap      = math.MaxUint8
+	acceptChSize  = 20
```
ayuryshev: Configuration value?

pkg/dmsg/frame.go
```diff
@@ -0,0 +1,160 @@
...
+	headerLen     = 5 // fType(1 byte), chID(2 byte), payLen(2 byte)
+)
+
+func isInitiatorID(tpID uint16) bool { return tpID%2 == 0 }
```
ayuryshev, line 34: Is it needed to give name for even-odd checks?

pkg/dmsg/frame.go
```diff
@@ -0,0 +1,160 @@
...
+
+func randID(initiator bool) uint16 {
+	var id uint16
+	for {
+		id = binary.BigEndian.Uint16(cipher.RandByte(2))
+		if initiator && id%2 == 0 || !initiator && id%2 != 0 {
```
ayuryshev, line 40: random value here is not required to be crypto-random, right?

Why not rand.IntN(math.MaxUInt16) ?

pkg/dmsg/frame.go
```diff
@@ -0,0 +1,160 @@
...
+			return id
+		}
+	}
+}
+
+var serveCount int64
+
+func incrementServeCount() int64 { return atomic.AddInt64(&serveCount, 1) }
```
ayuryshev, line 48: Are those functions needed?

pkg/dmsg/frame.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return names[ft]
+}
+
+// Frame types.
+const (
+	RequestType = FrameType(0x1)
+	AcceptType  = FrameType(0x2)
+	CloseType   = FrameType(0x3)
+	FwdType     = FrameType(0xa)
```
ayuryshev, line 75: Why `0xa`?

pkg/dmsg/frame.go
```diff
@@ -0,0 +1,160 @@
...
+	f = append(f, make([]byte, f.PayLen())...)
+	_, err := io.ReadFull(r, f[headerLen:])
+	return f, err
+}
+
+type writeError struct{ error }
+
+func (e *writeError) Error() string { return "write error: " + e.error.Error() }
+
+func isWriteError(err error) bool {
```
ayuryshev, line 133: Why this function needed?

pkg/dmsg/frame.go
```diff
@@ -0,0 +1,160 @@
...
+	_, ok := err.(*writeError)
+	return ok
+}
+
+func writeFrame(w io.Writer, f Frame) error {
```
ayuryshev, line 138: `writeFrame` here and below has meaning `sendPacket`?

pkg/dmsg/frame_test.go
```diff
@@ -0,0 +1,559 @@
...
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/skycoin/skywire/internal/ioutil"
+	"github.com/skycoin/skywire/pkg/cipher"
+)
+
+func Test_isInitiatorID(t *testing.T) {
```
ayuryshev, line 15: Is it test for even/odd?

pkg/dmsg/frame_test.go
```diff
@@ -0,0 +1,559 @@
...
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := isInitiatorID(tc.args.tpID)
+			assert.Equal(t, tc.want, got)
+		})
+	}
+}
+
+func Test_randID(t *testing.T) {
```
ayuryshev, line 45: and this too?

pkg/dmsg/frame_test.go
```diff
@@ -0,0 +1,559 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			got := randID(tc.args.initiator)
+			isEven := got%2 == 0
+
+			assert.Equal(t, tc.isEven, isEven)
+		})
+	}
+}
+
+func TestFrameType_String(t *testing.T) {
```
ayuryshev, line 77: I would consider to write it:

```golang
func ExampleFrameType_String() {
	fmt.Printf("%v\n", RequestType)
	fmt.Printf("%v\n", AcceptType)
	fmt.Printf("%v\n", CloseType)
	fmt.Printf("%v\n", FwdType)
	fmt.Printf("%v\n", AckType)
	fmt.Printf("%v\n", FrameType(0x0))
	fmt.Printf("%v\n", FrameType(0xFF))
	// Output: REQUEST
	// ACCEPT
	// CLOSE
	// FWD
	// ACK
	//
	// UNKNOWN:255
}
```

pkg/dmsg/frame_test.go
```diff
@@ -0,0 +1,559 @@
...
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := MakeFrame(tc.args.ft, tc.args.chID, tc.args.pay)
+			assert.Equal(t, tc.want, got)
+		})
+	}
+}
+
+func TestFrame_Type(t *testing.T) {
```
ayuryshev, line 177: Second test for converting FrameType?

pkg/dmsg/server.go
```diff
@@ -0,0 +1,368 @@
...
+	"github.com/skycoin/skywire/pkg/cipher"
+	"github.com/skycoin/skywire/pkg/messaging-discovery/client"
+)
+
+// ErrListenerAlreadyWrappedToNoise occurs when the provided net.Listener is already wrapped with noise.Listener
+var ErrListenerAlreadyWrappedToNoise = errors.New("listener is already wrapped to *noise.Listener")
+
+// NextConn provides information on the next connection.
+type NextConn struct {
+	l  *ServerConn
```
ayuryshev: `l *ServerConn`: 
- `l` is not recommended as variable name because of similarity with `1`
- why connection named `l`?

pkg/dmsg/server.go
```diff
@@ -0,0 +1,368 @@
...
+
+// ServerConn is a connection between a dmsg.Server and a dmsg.Client from a server's perspective.
+type ServerConn struct {
+	log *logging.Logger
+
+	net.Conn
+	remoteClient cipher.PubKey
+
+	nextRespID uint16
+	nextConns  [math.MaxUint16 + 1]*NextConn
```
ayuryshev, line 45: Please explain here.

A) There is another array of 65536 *NextConn created from start - what is the motivation for this?

B) Does it means ServerConn is recursive structure?
Something similar with:

```golang
type ServerConnTree struct {
	log *logging.Logger
	net.Conn
	remoteClient cipher.PubKey

	nextConns [65536]*ServerConnTree
}
```
Why?

pkg/dmsg/server.go
```diff
@@ -0,0 +1,368 @@
...
+
+		select {
+		case <-ctx.Done():
+			return 0, ctx.Err()
+		default:
+		}
+	}
+
+	id := c.nextRespID
+	c.nextRespID = id + 2
```
ayuryshev, line 91: Why not atomic AddUint?

pkg/dmsg/server.go
```diff
@@ -0,0 +1,368 @@
...
+	n := len(s.conns)
+	s.mx.RUnlock()
+	return n
+}
+
+// Close closes the dms_server.
+func (s *Server) Close() (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = errors.New("listener has not started")
```
ayuryshev: Please explain this error message

pkg/dmsg/server.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+	}()
+	if err = s.lis.Close(); err != nil {
+		return err
+	}
+
+	s.mx.Lock()
+	s.conns = make(map[cipher.PubKey]*ServerConn)
```
ayuryshev, line 301: Why it's needed to recreate s.conns on Close?

pkg/dmsg/server_test.go
```diff
@@ -0,0 +1,787 @@
...
+		name string
+		conn *ServerConn
+		ctx  context.Context
+		want want
+	}{
+		{
+			name: "ok",
+			conn: &ServerConn{
+				remoteClient: pk,
+				log:          logging.MustGetLogger("ServerConn"),
```
ayuryshev, line 57: I would consider add 
```golang
func TestingMain(t *testing.T){
}
```
and init repetetive things in it

pkg/dmsg/server_test.go
```diff
@@ -0,0 +1,787 @@
...
+	go s.Serve() //nolint:errcheck
+
+	time.Sleep(time.Second)
+
+	assert.NoError(t, s.Close())
+}
+
+// TestServer_Serve ensures that Server processes request frames and
+// instantiates transports properly.
+func TestServer_Serve(t *testing.T) {
```
ayuryshev, line 174: This function is ~400 LOCs
I would consider to split it

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,364 @@
...
+	select {
+	case <-tp.done:
+		return true
+	default:
+		return false
+	}
+}
+
+// Edges returns the local/remote edges of the transport (dms_client to dms_client).
+func (tp *Transport) Edges() [2]cipher.PubKey {
```
ayuryshev, line 122: I would consider to Sort Edges only in NewTransport
And then make them immutable

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// Type returns the transport type.
+func (tp *Transport) Type() string {
+	return Type
+}
+
+// Inject injects a frame from 'ClientConn' to transport.
+// Frame is then handled by 'tp.Serve'.
+func (tp *Transport) Inject(f Frame) error {
```
ayuryshev: What Inject means here? Send?

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,364 @@
...
+		tp.inMx.Unlock()
+	})
+
+	tp.serve() // just in case.
+	tp.ackWaiter.StopAll()
+	return closed
+}
+
+// Close closes the dmsg_tp.
+func (tp *Transport) Close() error {
```
ayuryshev, line 104: Is it safe to call Close() multiple times?

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,364 @@
...
+							tp.close()
+						}
+					}()
+				}
+
+				// add payload to 'buf'
+				tp.buf = append(tp.buf, p[2:])
+
+				// notify of new data via 'bufCh'
+				select {
```
ayuryshev, line 279: What is the meaning of `case <- tp.done:` here?

I see the only possible sense is to send `tp.bufCh <- struct{}{}:` asynchronously.

I suppose it will be equivalent to:
```golang
go func(tp *Transport){ tp.bufCh <- struct{}{} }(tp)
```

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,364 @@
...
+				tp.log.Infof("Rejected [%s]: Unexpected frame, possibly malicious server (ignored for now).", f.Type())
+			}
+		}
+	}
+}
+
+// Read implements io.Reader
+// TODO(evanlinjin): read deadline.
+func (tp *Transport) Read(p []byte) (n int, err error) {
+	<-tp.serving
```
ayuryshev, line 315: Does it needed to guard transport with `<-tp.serving`?

pkg/node/node.go
```diff
@@ -85,7 +85,7 @@ type PacketRouter interface {
 type Node struct {
 	config    *Config
 	router    PacketRouter
-	messenger *messaging.MsgFactory
+	messenger *dmsg.Client
```
ayuryshev, line 88: Why not `transport.Factory` here?

pkg/transport/handshake_test.go
```diff
@@ -88,87 +88,86 @@ func Example_newHsMock() {
...
-		})
-	}
-
-	sEntry := &SignedEntry{Entry: entry, Signatures: [2]cipher.Sig{}}
-	require.True(t, sEntry.Sign(pk1, sk1))
-	require.True(t, sEntry.Sign(pk2, sk2))
-
-	require.NoError(t, validateSignedEntry(sEntry, tr, pk1))
-}
+//func Example_validateEntry() {
```
ayuryshev, line 91: For removal or revival?

pkg/transport/transport.go
```diff
@@ -50,3 +52,35 @@ type Factory interface {
 	// Type returns the Transport type.
 	Type() string
 }
+
+// MakeTransportID generates uuid.UUID from pair of keys + type + public
+// Generated uuid is:
+// - always the same for a given pair
+// - GenTransportUUID(keyA,keyB) == GenTransportUUID(keyB, keyA)
+func MakeTransportID(keyA, keyB cipher.PubKey, tpType string, public bool) uuid.UUID {
```
ayuryshev, line 60: I thought about moving those funcs from `manager.go`
And I thought about moving them near transport.Entry types.

ayuryshev (commented):
1. Code is much more readable than in `messaging` and package itself much more simple
2. There are still some places that not so clear - explanations or changes needed
3. Overall - good work! It's a progress over messaging

pkg/dmsg/client.go
evanlinjin: https://talks.golang.org/2014/names.slide#8

internal/noise/noise.go
evanlinjin: I personally don't see this as a problem. Might need some input from others.

internal/ioutil/buf_read.go
evanlinjin: @ayuryshev I think we should leave it as it. We may use the same logic again for later implementations of `io.Reader`

Makefile
evanlinjin: We will eventually move all `dmsg`-associated parts to it's own repo. It will be a good call to move this here first.

internal/ioutil/atomic_bool.go
evanlinjin: I think we can leave it for now. We may use it later.

pkg/dmsg/README.md
evanlinjin: No particular reason. I just wanted to separate FWD and ACK frame types from the rest.

pkg/dmsg/client.go
evanlinjin: 1. I think in this context it means "transports".
2. We either use maps for this. This is more performant (but uses more memory). This needs to be performant, as we call from it every time a frame is received.

pkg/dmsg/client.go
evanlinjin: Because operations to the associated `uint16` value needs to be atomic in combination with other operations. Hence, `atomic` will not be a viable solution.

pkg/dmsg/client.go
evanlinjin: Not a viable solution.

pkg/dmsg/client.go
evanlinjin: Yes, as we use `sync.Once`

pkg/dmsg/client.go
evanlinjin: Yes. The meaning of this error is that the connection is already closed. I was following the behavior of how `.Close` would function in `net` interfaces. But it is not needed.

pkg/dmsg/client.go
evanlinjin: It's a matter of taste. I like this better.

pkg/dmsg/client.go
evanlinjin: Correct.

pkg/dmsg/client.go
evanlinjin: `Conn` was called `Link`before.

pkg/dmsg/client.go
evanlinjin: @ayuryshev generally I would agree. But I think the readability is fine here.

pkg/dmsg/client_test.go
evanlinjin: I think it is fine as is. As long as the test is easy to read and does what is should, I don't think it is worth changing.

pkg/dmsg/client_test.go
evanlinjin: It's so that we can directly get two `ClientConn`s communicating with one another.

evanlinjin: I'll write a comment.

internal/noise/noise.go
ayuryshev: It's not a problem
But:
- pkg/log - from stdlib
- var log in util/logging
- and here `log` again 

All three places it has different meaning: package, MasterLogger, PackageLogger

pkg/dmsg/client_test.go
evanlinjin: Yes. Comment added.

internal/ioutil/atomic_bool.go
ayuryshev: In this case I would consider to change signature to:
```golang
func (b *AtomicBool) Set(v bool) 
```
detection "if the value is changed" - is simpler to write with CompareAndSwap when needed

internal/ioutil/buf_read.go
ayuryshev: I don't understand how it helps to write `Reader`s

pkg/dmsg/frame.go
evanlinjin: It makes the code more readable.

pkg/dmsg/frame.go
evanlinjin: I think it is fine.

pkg/dmsg/frame.go
evanlinjin: Yes. Those are used.

pkg/dmsg/frame.go
evanlinjin: For convenience.

pkg/dmsg/frame.go
evanlinjin: We are calling the data units of `dmsg` frames.

pkg/dmsg/frame_test.go
evanlinjin: I wouldn't use an example for it, but I think the use of `cases` is a little excessive.

pkg/dmsg/frame_test.go
evanlinjin: Yes, I do agree that the test for frames is a little excessive.

pkg/dmsg/server.go
evanlinjin: A) Performance > Memory usage.
B) I wouldn't call it a tree structure. The Server stores connections with Clients, a frame sent from one client needs to get forwarded to another client.

pkg/dmsg/server.go
evanlinjin: The `AddUint` and other operations together (in a group) needs to be handled atomically. So not just the `uint16` value itself.

internal/noise/noise.go
evanlinjin: I think because of the context of where we are using it, it is fine.

pkg/dmsg/server.go
evanlinjin: Legacy code. removed.

pkg/dmsg/server.go
evanlinjin: @ayuryshev just in case some goroutines try to read/write to it after close is called.

pkg/dmsg/server_test.go
evanlinjin: Consider each sub-test as it's own function. Yes some sub-tests needs simplifications.

pkg/dmsg/transport.go
evanlinjin: In the future, `transport.Transport` interface will have `Local() cipher.Pubkey` and `Remote() cipher.Pubkey` definitions. The `Edges()` addition turned out to be a hassle in certain cases (because we need additional calls to determine which is the remote/local edge). I think we can leave it for now.

pkg/dmsg/transport.go
evanlinjin: add, introduce, insert, instill, etc.

evanlinjin: I'm happy with this naming convention. Also open for other opinions.

pkg/dmsg/transport.go
evanlinjin: Yes

pkg/dmsg/transport.go
evanlinjin: > What is the meaning of `case <- tp.done:` here?

Return on close.

> I see the only possible sense is to send `tp.bufCh <- struct{}{}:` asynchronously.

The chan is buffered with size of 1.

> I suppose it will be equivalent to: ...

No, this introduces unnecessary go routines.

pkg/node/node.go
evanlinjin: @ayuryshev I think we can leave it for now. It's for calling the `InitiateServerConnections` function.

pkg/transport/handshake_test.go
evanlinjin: Revival

pkg/dmsg/README.md
evanlinjin: I think entity is fine. We talk about `skywire services` so I didn't want to reuse "services". "Service" also suggests running as an executable on a machine, but it is not necessarily the case here.

pkg/dmsg/README.md
evanlinjin: Yes.

pkg/dmsg/README.md
evanlinjin: `dmsg` Entity. They are both discoveries (dmsg discovery, transport discovery), discoveries have entries.

pkg/dmsg/README.md
evanlinjin: The point here is that it is represented by a number. We mention later on about using the even/odd property of the ID to determine role.

pkg/dmsg/README.md
evanlinjin: "line of communication" and "lines of communication" is common English. It's not military-specific.

ayuryshev: It's not network terminology either

pkg/dmsg/README.md
evanlinjin: We are using `frame` as the term for data units within `dmsg`.

pkg/dmsg/README.md
evanlinjin: We have the definition: `**transport ID -** A uint16 value that identifies a transport.`

You can say that it has only internal meaning for dmsg, but if we want to ignore all internal meanings, then the definition for `transport.Transport` interface will be sufficient enough for `dmsg` specs.

pkg/dmsg/README.md
evanlinjin: We state in `TODO`: ``ACK` frames should include the first 4 bytes of the rolling hash of incoming payloads, enforcing reliability of data. Transports should therefore keep track of incoming/outgoing rolling hashes.`

and yes.

pkg/dmsg/client.go
ayuryshev: I would consider map here - exactly as in comment.
Performance here will be all eaten with network io and concurrency.

pkg/dmsg/transport.go
evanlinjin: Yes. We need to wait until the transport handshake completes.

pkg/dmsg/transport.go
ayuryshev: `Inject` - has strong connotation with packet injection.
It seems clear that it was not intentional.
`Transport.SendFrame`? `Transport.SendPacket`?

I prefer `Packets`: `Frames` usually for L2 and we are not doing L2 now.

pkg/dmsg/transport.go
ayuryshev: There is no return on `case <-tp.done:`
And golang guarantee randomness of `select-case` branches.
So this construction will work completely random after `close(tp.done)`

Correct if I'm wrong here.

pkg/node/node.go
ayuryshev: It's not a problem at all:

```golang
func (node *Node) Start() error {
	ctx := context.Background()
	switch msgr := node.messenger.(type) {
	case *dmsg.Client:
		err := msgr.InitiateServerConnections(ctx, node.config.Messaging.ServerCount)
		if err != nil {
			return fmt.Errorf("%s: %s", dmsg.Type, err)
		}
		node.logger.Info("Connected to dmsg servers")
	case *messaging.MsgFactory:
		err := msgr.ConnectToInitialServers(ctx, node.config.Messaging.ServerCount)
		if err != nil {
			return fmt.Errorf("ConnectToInitialServers: %s", err)
		}
		node.logger.Info("Connected to messaging servers")
	}
```

pkg/transport/handshake_test.go
ayuryshev: OK

pkg/dmsg/transport.go
evanlinjin: There is a possibility that the `tp.done` is going to be selected, but I do agree that it's not needed as we have the `default:` case.

Note that the `default` case is always selected last.

pkg/node/node.go
evanlinjin: @ayuryshev ideally that function is to be called outside of `Node`. But it's not. The solution you have provided is elegant, but it's not the optimal way.

I would rather stick with what's in the PR before we change to call `Initiate...` outside of `Node`.

pkg/dmsg/README.md
ayuryshev: I mean: details of even/odd, positive/negative, uint16/uint32 - could be absolutely insignificant if it's only dmgs implementation details.

We don't even need to document this if it's not public API

pkg/dmsg/README.md
ayuryshev: So it's packets of data?

evanlinjin: The term we are using for the data unit within `dmsg` is `frame`. `packet` is for `Router`.

pkg/dmsg/README.md
evanlinjin: I think these provided details allow the codebase to be more readable. I will leave it as is.

pkg/dmsg/client.go
evanlinjin: For sake of convenience, I will create a separate issue for this.

pkg/dmsg/client.go
ayuryshev: Please explain.
What's needed from nextInitID?

`atomic.AddUint32` for increment
and
`atomic.LoadUint32` for get value

What else is needed?

pkg/dmsg/client.go
ayuryshev: It's a good talk
And `b []byte` - indeed is traditional but `p []byte` - not.
I would consider to add some symbols here:
- `data`
- `frameData`
- `payload`
- `packedData`

Looks OK.

pkg/dmsg/client.go
evanlinjin: The increment_next_id and get_next_id operations (together) need to be atomic. Otherwise we may run into the possibility of trying to allocate the same transport id in two different transports.

pkg/dmsg/client.go
ayuryshev: Mutability is not a matter of taste. 
Each mutable member requires attention.
I could not imagine situation when it's needed to change Logger during Client lifetime.

I would consider to:
- init logger once
- remove SetLogger

pkg/dmsg/client.go
ayuryshev: Thanks for explanation

pkg/dmsg/client.go
ayuryshev: It's not about readability but semantics. 
We had something similar in `transport.Manager` when critical code was unnamed.
IMO this code is equally significant - it deserves to at least be named.

pkg/dmsg/client.go
evanlinjin: For note in the future: "I think it's better if ... because ..." would be a better tone of voice and provide more reasoning when compared with "why ...?".

You have a point. I'll change this.

pkg/dmsg/client_test.go
ayuryshev: This test is easy to beatify with simple deduplication.
Look: https://github.com/skycoin/skywire/blob/35e3bcb7c21b5b701823b3be57cfe6db4ead3c2d/pkg/dmsg/client_test.go#L209

I did not finished but idea must be clear:

```golang
		errCh := closeClosers(
			tr1, tr2, conn1, conn2, conn3, conn4,
		)

		assertNoError(t, errCh)

		assertFalse(t,
			isDoneChannelOpen(conn1.done),
			isDoneChannelOpen(conn3.done),
			isDoneChannelOpen(tr1.done),
			isReadChannelOpen(tr1.inCh),
			isDoneChannelOpen(tr2.done),
			isReadChannelOpen(tr2.inCh),
		)
```

pkg/dmsg/frame.go
ayuryshev: I tend to disagree.
Simple task is solved in a complex way. 

Look:

```golang
func randID(initiator bool) uint16 {
	id := rand.Intn(tpBufCap)
	if id%2 == 0 && initiator {
		return uint16(id)
	}
	return uint16(id + 1)
}
```

pkg/dmsg/frame.go
ayuryshev: I  mean: why envelop just one call atomic.AddInt64 into function?
It's simple enough and absolutely readable

pkg/dmsg/frame_test.go
ayuryshev: `Example` is perfectly fits here: it's a test for `String()` representation.

pkg/dmsg/server.go
ayuryshev: Look:

```golang
type NextConn struct {
	l  *ServerConn
	id uint16
}

type ServerConn struct {
	nextConns  [math.MaxUint16 + 1]*NextConn
}
```
Those types are mutually recursive.

And I don't see `Client`s here - there are only `ServerConn`s here.
What is the motivation by this?

pkg/dmsg/server.go
ayuryshev: OK

pkg/dmsg/server.go
evanlinjin: @ayuryshev 
Yes, you can say that it is a recursive structure, however it is still not a tree.

> And I don't see Clients here - there are only ServerConns here.
> What is the motivation by this?

Why do we need to have `Client` here? Maybe I've misunderstood the question?

pkg/dmsg/transport.go
```diff
@@ -128,9 +128,8 @@ func (tp *Transport) Type() string {
 	return Type
 }
 
-// Inject injects a frame from 'ClientConn' to transport.
```
ayuryshev: My comment is not for this line but for line 98:
```golang
	tp.serve() // just in case.
	tp.ackWaiter.StopAll()
	return closed
}
```
(for reasons unknown for me github prohibits writing comment  directly)

I suppose that "just in case" - could be a subtle error.

evanlinjin: @ayuryshev what kind of subtle error are you suggesting?
- The actions within `tp.serve()` are protected with `sync.Once`.
- We need to call `tp.serve()` to unblock Read/Write (in case `tp.Serve` is not called).

pkg/dmsg/frame_test.go
evanlinjin: https://github.com/skycoin/skywire/issues/420

pkg/dmsg/server_test.go
evanlinjin: https://github.com/skycoin/skywire/issues/420

pkg/dmsg/server_test.go
evanlinjin: https://github.com/skycoin/skywire/issues/420

pkg/dmsg/frame_test.go
evanlinjin: https://github.com/skycoin/skywire/issues/420

pkg/dmsg/client.go
nkryuchkov: What do you think of implementing a benchmark to compare array and map in this case?

evanlinjin: @nkryuchkov good idea, however this would be rather a time-consuming task because we need to consider multiple scenarios. I've created a ticket: https://github.com/skycoin/skywire/issues/419

Let's continue the discussion there!

pkg/dmsg/frame.go
nkryuchkov: @ayuryshev it's easier to read and abstracts the way how the value should be changed

pkg/dmsg/frame_test.go
nkryuchkov: @ayuryshev it tests that checking if an ID is initiator works correctly

pkg/dmsg/frame_test.go
nkryuchkov: @ayuryshev It tests that an ID is generated correctly depending on if we need an initiator one or not.

evanlinjin: @ayuryshev I'm sure that it is obvious to you how the test works. If you think the test can be improved, it is better to write that in a statement detailing your thoughts.

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,362 @@
...
+type Transport struct {
+	net.Conn // underlying connection to dmsg.Server
+	log      *logging.Logger
+
+	id     uint16        // tp ID that identifies this dmsg.Transport
+	local  cipher.PubKey // local PK
+	remote cipher.PubKey // remote PK
+
+	inCh chan Frame   // handles incoming frames (from dmsg.Client)
+	inMx sync.RWMutex // protects 'inCh'
```
ayuryshev, line 36: This mutex  is not needed.
It's used in `Transport.HandleFrame` and used incorrectly to remedy incorrect `select` statement.

After changing HandleFrame into:

```golang
func (tp *Transport) HandleFrame(f Frame) error {
	if tp.IsClosed() {
		return io.ErrClosedPipe
	}

	tp.inCh <- f
	return nil
```
Everything works

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,362 @@
...
+	return transport.SortPubKeys(tp.local, tp.remote)
+}
+
+// Type returns the transport type.
+func (tp *Transport) Type() string {
+	return Type
+}
+
+// HandleFrame allows 'tp.Serve' to handle the frame (typically from 'ClientConn').
+func (tp *Transport) HandleFrame(f Frame) error {
```
ayuryshev, line 132: Look at my comment about `Transport.inMx`

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,362 @@
...
+	defer func() {
+		if err != nil {
+			tp.log.WithError(err).WithField("remote", tp.remote).Warnln("(HANDSHAKE) Rejected by remote.")
+		} else {
+			tp.log.WithField("remote", tp.remote).Infoln("(HANDSHAKE) Accepted by remote.")
+		}
+	}()
+
+	select {
+	case <-tp.done:
```
ayuryshev, line 189: I think it would be more safe to remove:
```golang
 case <-tp.done:
   return
```

and substitute it with check before `select`:

```golang
if tp.IsClosed() {
   return io.ErrClosedPipe
}
```

pkg/dmsg/transport.go
```diff
@@ -0,0 +1,362 @@
...
+	// also write CLOSE frame if this is the first time 'close' is triggered
+	defer func() {
+		if tp.close() {
+			_ = writeCloseFrame(tp.Conn, tp.id, 0) //nolint:errcheck
+		}
+	}()
+
+	for {
+		select {
+		case <-tp.done:
```
ayuryshev, line 245: And here check for `case <-tp.done` would be better substituted with `if tp.IsClosed()`

pkg/dmsg/transport.go
evanlinjin: The mutex is needed because `close(tp.inCh)` and `tp.inCh <- f` operations are not thread-safe.

"Everything works" is not good enough. You cannot ensure, with this statement, that you have tested all scenarios and conditions.

However, you can definitely say:
- I've noticed that doing the following changes still allows the system to work.
- I feel the `select` statement here is faulty, because we cannot guarantee that the `done` case will be selected when it's given that `done` is closed.

pkg/dmsg/transport.go
evanlinjin: This comment is not constructive. I will resolve this thread.

pkg/dmsg/transport.go
evanlinjin: @ayuryshev please review my most recent changes as I feel it is a more elegant solution: https://github.com/skycoin/skywire/pull/402/commits/e9e11d025b147d07d3d84d3472612cf564ed414c

pkg/dmsg/transport.go
evanlinjin: As aforementioned, please check out my latest changes: https://github.com/skycoin/skywire/commit/e9e11d025b147d07d3d84d3472612cf564ed414c

Issue #388, opened 2019-06-04
Feature: implement a reliable method to kill apps spawned by a specific skywire-node


ayuryshev:
**Feature description**

Create a reliable method to kill apps spawned by specific skywire-node

**Is your feature request related to a problem? Please describe.**

Concider the following scenario:
- 3 nodes are running 
- NodeA and NodeC both spawns skychat
- NodeB is intermediary

```bash 
$ ps aux |grep "[N]ode"                                                                                                                             
# isk      10310  6.3  0.4 1451492 33184 pts/7   Sl+  14:03   0:01 ./bin/skywire-node ./integration/messaging/nodeA.json --tag NodeA
# isk      10476 10.7  0.4 1377528 35620 pts/8   Sl+  14:03   0:01 ./bin/skywire-node ./integration/intermediary-nodeB.json --tag NodeB
# isk      10634  7.2  0.4 1377188 33424 pts/9   Sl+  14:03   0:01 ./bin/skywire-node ./integration/messaging/nodeC.json --tag NodeC
$ ps aux |grep "[s]kychat"                                                                                                                          
isk      10444  1.3  0.1 839896 15212 pts/7    Sl+  14:03   0:00 /home/isk/github.com/skycoin/src/skywire-services/bin/apps/skychat.v1.0
isk      10710  1.5  0.1 838492 15492 pts/9    Sl+  14:03   0:00 /home/isk/github.com/skycoin/src/skywire-services/bin/apps/skychat.v1.0 -addr :8001
```
Which one skychat is spawned by which node?

This problem is especially important for implementing integration tests with restart cycles: hanged apps makes impossible to start them reliably.

**Describe the solution you'd like**

Solution that allows to reliably kill apps spawned by specific node without killing apps spawned by the other nodes.

**Describe alternatives you've considered**

- Improve skywire-node app-management in such way that it kills it's apps during startup 
- Improve apps: add some kind of health-check with heartbeats so that apps automatically exits when skywire-node not exists 

**Possible implementation**

1. Use `--tag` param of skywire-node for apps too 

This way it will be possible to distinguish spawned apps with `ps aux|grep "[t]agName"`

2. Use of pid-files

Propose such solution in comments please

evanlinjin:
I've made this clear, but I would choose option: `2. Use of pid-files`. Unless if @ayuryshev has some better ideas?

ayuryshev:
Solution with pid-files residing in local directory of nodes - I agree, it will work.

[...]

PR #403, opened 2019-06-10
added pid files. permission denied error

ivcosla:


ivcosla:
Closes #388

ayuryshev:
So how this PR is intended to work?
What becomes if skywire-node panics with unrecovered error?
What will be on restart of this skywire-node?
Is there a recipe to detect hanged apps without node?

ivcosla:
@ayuryshev 
1) When apps are spawned their PIDS are persisted in disk. 
2) When node is started, before it spawns apps, it reads such file and kills every app previously registered, then clears the pid file. Then apps are spawned and we go back to step 1).

If a node panics, when it is restarted it should kill the apps before it spawns the new ones, so app-node communication will work.

The only way to test it that I had was to run integration environments and use apps until one of them hangs on node close. You can check that by using `ps aux| grep [app name]`, then you can start the node again and check in the logs that the new app spawned correctly, if you perform the above command again you will see now such app has a different PID.

You can also check how every time that you start a node there is an `apps.pid` file inside `~/.skycoin/skywire/[node pk]/` directory with the name and pid of every node-spawned app. On every run it's pids must be different.

ayuryshev:
I just today found a reproducible way to panic node leaving hanged  apps:

in skywire-services: 
1. run-messaging-env.sh
2. start messaging
3. kill messaging-server

I see that NodeA and NodeC panics 
and leave skychat running
must be good for testing this PR

pkg/node/node.go
```diff
@@ -239,6 +249,61 @@ func (node *Node) Start() error {
 	return nil
 }
 
+func (node *Node) dir() string {
+	return pathutil.NodeDir(node.config.Node.StaticPubKey)
+}
+
+func (node *Node) pidFile() *os.File {
+	f, err := os.OpenFile(filepath.Join(node.dir(), "apps.pid"), os.O_RDWR|os.O_CREATE, 0600)
```
ayuryshev: 0644 maybe? To give ability to read for not-owners

pkg/node/node.go
```diff
@@ -389,6 +459,14 @@ func (node *Node) SpawnApp(config *AppConfig, startCh chan<- struct{}) error {
 	return appErr
 }
 
+func (node *Node) persistPID(name string, pid int) {
```
ayuryshev, line 462: Do I understand correctly that content of pid-file is:
app-name pid

?

ayuryshev (commented):
1. File permissions 0750 and 0600 (for directories and files) -  why not 755 and 0644 (typically used for pidfiles) ?
Examples:
- https://github.com/facebookarchive/pidfile/blob/f242e2999868dcd267a2b86e49ce1f9cf9e15b16/pidfile.go#L48
- https://github.com/moby/moby/blob/349d4dd0c981232a95fc178563a54792699d2823/pkg/pidfile/pidfile.go#L40

2. Did I understood correctly that one pid per node contains multiple lines one per application?

pkg/node/node.go
ivcosla: yes

pkg/node/node.go
ivcosla: it was 0644 before, but go-sec linter complained about it and pointed me to use 0600 instead.

evanlinjin: lol, guess 0600 is safer anyhow hahaha.

evanlinjin (approved):
Looks good.

Issue #409, opened 2019-06-13
Make apps retry on network error


ivcosla:
Leverage that responsibility on apps, instead of re-create transports at messaging level.

This can potentially work around some issues on messaging and save time on performing handshakes.

jdknives:
Closed by #414

PR #414, opened 2019-06-15
Feature/apps retry on network error

ivcosla:
Attempt to improve resiliency on apps. #409


internal/netutil/retrier.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// WithErrWhitelist sets a list of errors into the retrier, if the RetryFunc provided to Do() fails with one of them it will return inmediatelly with such error
+func (r *Retrier) WithErrWhitelist(errors ...error) *Retrier {
+	m := make(map[error]struct{})
+	for _, err := range errors {
+		m[err] = struct{}{}
+	}
+
+	r.errWhitelist = m
```
Darkren, line 44: Wouldn't it be good to protect it with a mutex? Just in case it gets called somewhere in goroutines

Darkren (commented):
Looks great to me, just a small question arised

internal/netutil/retrier.go
ivcosla: Well, this is supposed to be called only when creating the retrier: `NewRetrier(...).WithErrWhiteList(...)`, so afterwards concurrent access to it will only be for read operations.

It was supposed to look like a fluent interface, however I didn't implemented two types (one `retrierBuilder` and one `retrier`), so now ones can call `WithWhiteListErr` on the retrier and make a bad use of it.

The options that I see:
 - implement both types (https://github.com/lann/builder for reference), but for the amount of fields that this struct has I don't like it
 - add a mux
 - remove `WithWhiteListErr()` and add another parameter: `NewRetier(..., errWhiteList ...error)`
 - document the function warning of just using it for initialization or at least don't calling it again concurrently

I prefer the last one. Opinion on that?

cmd/apps/skychat/chat.go
```diff
@@ -109,7 +112,10 @@ func messageHandler(w http.ResponseWriter, req *http.Request) {
 
 	if conn == nil {
 		var err error
-		conn, err = chatApp.Dial(addr)
+		err = r.Do(func() error {
```
evanlinjin, line 115: I'm not sure if using the retry logic with `skychat` is going to be beneficial in any way. The main intention for the retry logic (from my understanding anyway) is to stop apps from quitting when the network fails. From my understanding, `skychat` doesn't quit on network failure? And on every send request, if there is no "connection", `skymsg` tries to establish a connection?

Let me know if I'm mistaken, or if I've missed something.

internal/netutil/retrier.go
evanlinjin: I guess if @ivcosla made a comment (somewhere) that said that `Retrier` is not thread-safe, then it is fine.

internal/therealproxy/client.go
```diff
@@ -40,7 +48,10 @@ func (c *Client) ListenAndServe(addr string) error {
 			return fmt.Errorf("accept: %s", err)
 		}
 
-		stream, err := c.session.Open()
+		err = r.Do(func() error {
```
evanlinjin: I'm not sure if this is helpful here... If there is something wrong with a connection, it won't *heal*. We need to re-accept.

pkg/app/app.go
```diff
@@ -130,12 +130,14 @@ func (app *App) Dial(raddr *Addr) (net.Conn, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	fmt.Println("loop created")
```
evanlinjin: Might need to get rid of these log messages.

evanlinjin (commented):
`Retrier` is beautifully written.

The problem right now is that it is hard to determine where best to put the retry logic. Our goal is to have apps only quit when these following events occur:
1. When the parent node quits.
2. When the user tells the app to quit.

I'm not sure whether this PR demonstrates this. We need to give this more thought.

pkg/app/app.go
ivcosla: damn, missed that one

evanlinjin:
> `Retrier` is beautifully written.
> 
> The problem right now is that it is hard to determine where best to put the retry logic. Our goal is to have apps only quit when these following events occur:
> 
>     1. When the parent node quits.
> 
>     2. When the user tells the app to quit.
> 
> 
> I'm not sure whether this PR demonstrates this. We need to give this more thought.

I take some of this back. It seems, after mild testing that it really does stop apps from quitting. I'm not sure exactly how. But please review the comments as maybe we remove some unnecessary code.

cmd/apps/skychat/chat.go
ivcosla: I think is of no harm, basically, I'm trying to retry when adquiring a connection. So if there is a small problem with latency for example, it can recover. But you are right, skychat doesn't fails on network errors, except for writing on closed channels, that I added a return in the same file

internal/therealproxy/client.go
ivcosla: I thought this was a "Dial" operation, so I also tried to retry it. However I think that the retry at the app level of proxy should be enough to fix at least our biggest problem with it (the app exit when it doesn't finds a route)

internal/netutil/retrier.go
evanlinjin: @ivcosla last one definitely.

cmd/apps/skychat/chat.go
```diff
@@ -122,10 +128,18 @@ func messageHandler(w http.ResponseWriter, req *http.Request) {
 		go handleConn(conn)
 	}
 
-	if _, err := conn.Write([]byte(data["message"])); err != nil {
+	err := r.Do(func() error {
```
evanlinjin: Let's not retry here. The connection does not heal itself once broken.

cmd/apps/skychat/chat.go
evanlinjin: Fine. Let's leave `Dial` in a retry loop here.

evanlinjin (approved):

Issue #404, opened 2019-06-12
Bug: Incorrect syslog hook in skywire-nod


ayuryshev:
**Describe the bug**

skywire-node creates logrus.Logger multiple times for every subsystem.

Hooks added by logging.AddHook in cmd/skywire-node/commands/root.go works only on upper level.

**Environment information:**

In every environment.

**Steps to Reproduce**

1. Run syslog-ng
2. Run skywire-node with --syslog param

E.g.
```bash
./skywire-node --syslog localhost:514 --tag NodeA
```

**Actual behavior**

Look at logs collected by syslog.
Notice that only logs from cmd/skywire-node/commands/root.go  were collected.

**Expected behavior**

Logs from all subsystems are expected.

**Additional context**

**Possible implementation**

Instead of `logging.MustGetLogger` each time in every package:

1. Create one `logging.NewMasterLogger` for skywire-node
2. Add hooks for MasterLogger
2. Create additional `PackageLogger` when needed

jdknives:
Closed bz #416

PR #416, opened 2019-06-19
Logging improvements for running and testing skywire-node

ayuryshev:
Logging improvements for running and testing skywire-node

This PR is based on #399

1. Fixed logging for skywire-node: there is only one MasterLogger creation now, fixed hooks on skywire-node level
2. Logging during tests: turned off by default, turn on by $TEST_LOGGING_LEVEL
3. Integragration tests customizable by $SYSLOG_OPTS variable
4. Makefile goal: `make run-syslog` - creates clean instance of syslog-ng with logs mounted to /tmp/syslog
5. README.md - changes documented with examples of usage
    
Closes: #404


cmd/skywire-node/commands/root.go
```diff
@@ -16,97 +18,173 @@ import (
 	"github.com/skycoin/skycoin/src/util/logging"
 	"github.com/spf13/cobra"
 
+	"net/http"
```
nkryuchkov, line 21: Could you please move this import and the below one up to the first import block in order to have them all in one block?

cmd/skywire-node/commands/root.go
```diff
@@ -16,97 +18,173 @@ import (
...
+		option = profile.CPUProfile
+	case "mem":
+		option = profile.MemProfile
+	case "mutex":
+		option = profile.MutexProfile
+	case "block":
+		option = profile.BlockProfile
+	case "trace":
+		option = profile.TraceProfile
+	default:
```
nkryuchkov: Could you consider removing empty `default` case here?

cmd/skywire-node/commands/root.go
```diff
@@ -16,97 +18,173 @@ import (
...
+	cfg.masterLogger = logging.NewMasterLogger()
+	cfg.logger = cfg.masterLogger.PackageLogger(cfg.tag)
 
-		conf := &node.Config{}
-		if err := json.NewDecoder(rdr).Decode(&conf); err != nil {
-			logger.Fatalf("Failed to decode %s: %s", rdr, err)
+	if cfg.syslogAddr != "none" {
+		hook, err := logrus_syslog.NewSyslogHook("udp", cfg.syslogAddr, syslog.LOG_INFO, cfg.tag)
+		if err != nil {
+			cfg.logger.Error("Unable to connect to syslog daemon")
```
nkryuchkov: I'd also log the returned error here

cmd/skywire-node/commands/root.go
```diff
@@ -16,97 +18,173 @@ import (
...
-				logger.Fatalf("Failed to open config: %s", err)
-			}
-		} else {
-			logger.Info("Reading config from STDIN")
-			rdr = bufio.NewReader(os.Stdin)
-		}
+func (cfg *runCfg) startProfiler() *runCfg {
+	var option func(*profile.Profile)
+	switch cfg.profileMode {
+	case "none":
```
nkryuchkov, line 85: What do you think about creating a `const` block and probably a new type with all possible values for `profileMode` in order to avoid comparison with string literals?

cmd/skywire-node/commands/root.go
```diff
@@ -16,97 +18,173 @@ import (
...
+	defer cfg.profileStop()
+	if err := cfg.node.Close(); err != nil {
+		if !strings.Contains(err.Error(), "closed") {
+			cfg.logger.Fatal("Failed to close node: ", err)
+		}
 	}
+	return cfg
+}
+
+// waitOsSignalsAndDelay
```
nkryuchkov: Is there any purpose for this comment?

cmd/skywire-node/commands/root.go
```diff
@@ -16,97 +18,173 @@ import (
...
+func (cfg *runCfg) waitOsSignals() *runCfg {
+	ch := make(chan os.Signal, 2)
+	signal.Notify(ch, []os.Signal{syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT}...)
+	<-ch
```
nkryuchkov, line 178: Could you please explain in short why reading from `ch` is used here?

internal/httpauth/client_test.go
```diff
@@ -6,14 +6,17 @@ import (
...
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/skycoin/skycoin/src/util/logging"
```
nkryuchkov, line 19: What do you think of swapping this line and the above line because `skycoin/skycoin` is another repository?

nkryuchkov (approved):
Looks good to me. Could you please consider making changes to the lines where I left comments?

cmd/skywire-node/commands/root.go
ayuryshev: If I was a developer of pkg/profile - I would concider this way.
But that's part of theirs API and recommended way to use this package.
Look at https://github.com/pkg/profile/blob/f6fe06335df110bcf1ed6d4e852b760bfc15beee/example_test.go#L42

IMO: API provide by package is "good enough". I don't see the point to improve it inside skywire-node.

cmd/skywire-node/commands/root.go
ayuryshev: Indeed. I forgot `make format` before push.
But not for this file. I never format by hand import in golang sources - it's too fragile.

cmd/skywire-node/commands/root.go
ayuryshev: Yes

cmd/skywire-node/commands/root.go
ayuryshev: It's a common patern in using os.signal:
- register signals 
- block execution waiting for signal
Look at examples https://golang.org/pkg/os/signal/#Notify

As for `make(chan os.Signal, 2)` - buffer size=2 is inherited. 
Maybe `1` will be sufficient. Not tested

internal/httpauth/client_test.go
ayuryshev: I pushed code after `make format`. 
It could be outdated or not - IMO such things must be solved automatically

cmd/skywire-node/commands/root.go
ayuryshev: Agree

cmd/skywire-node/commands/root.go
ayuryshev: removed

cmd/skywire-node/commands/root.go
nkryuchkov: @ayuryshev Yeah, it's a common pattern. I just don't fully understand why it happens before starting a goroutine, which tries to read from the channel again. Could this line be removed here and the code in the `select` below be unwrapped from the goroutine in order that the timeout works? 

However, I might incorrectly understand this code.

ayuryshev: It starts goroutine that allow skywire-node to shutdown in `cfg.conf.ShutdownTimeout` period of time or make it possible to kill skywire-node with second signal.

So it makes possible: "Ctrl-C [wait for shutdown]" or "Ctrl-C Ctrl-C"

nkryuchkov: @ayuryshev thanks for the explanation. I misunderstood the purpose of the method.

ivcosla (approved):
This is looking good to me

Issue #420, opened 2019-06-20
Improve dmsg tests.


evanlinjin:
The following are comments regarding `dmsg` test improvements. Do not consider then as the model implementation. But note that the queries need to be resolved in some form or another:

- [ ] https://github.com/skycoin/skywire/pull/402/files/81eee395f83803e10e8fb30d49790978f8215dee?file-filters%5B%5D=#r295770293

- [ ] https://github.com/skycoin/skywire/pull/402/files/81eee395f83803e10e8fb30d49790978f8215dee?file-filters%5B%5D=#r295770731

- [ ] https://github.com/skycoin/skywire/pull/402/files/81eee395f83803e10e8fb30d49790978f8215dee?file-filters%5B%5D=#r295776402

- [ ] https://github.com/skycoin/skywire/pull/402/files/81eee395f83803e10e8fb30d49790978f8215dee?file-filters%5B%5D=#r295778074

nkryuchkov:
To be done:

- [ ] https://github.com/skycoin/skywire/pull/423#discussion_r297051269

nkryuchkov:
Moved to https://github.com/skycoin/dmsg/issues/1

PR #423, opened 2019-06-24
DMSG reconnection tests

nkryuchkov:
Moved from https://github.com/evanlinjin/skywire/pull/8.

The tests have some space for improvements that will be done in https://github.com/skycoin/skywire/issues/420. For example, there're some code snippets duplicated among the tests that may be done in separate functions. Due to the requirement of modification of all tests, such code is left as it is. It would be appreciated if improvement proposals for reconnection tests only would be left here. For common improvement proposals, please leave comments in the related issue.


pkg/dmsg/server_test.go
```diff
@@ -767,6 +765,115 @@ func TestServer_Serve(t *testing.T) {
...
+		})
+
+		t.Run("Random address", func(t *testing.T) {
+			t.Parallel()
+			testReconnect(t, true)
+		})
+	})
+}
+
+func testReconnect(t *testing.T, randomAddr bool) {
```
evanlinjin, line 906: Looks good, however some comments describing what is being tested will be good. Also in regards to naming convention; initiator should go with responder, and local should go with remote.

evanlinjin (approved):

Darkren (approved):
Looks good to me

ivcosla:
By the travis logs it looks like the newly added tests are producing a lot of log messages. Logging should be turned off before merging this branch.

At the moment there is a conflict that needs to be fixed too.

pkg/dmsg/server_test.go
nkryuchkov: @evanlinjin Thanks for your review. I'll do this in https://github.com/skycoin/skywire/issues/420

PR #427, opened 2019-06-25
Cleanup logging for DMSG

Darkren:
Done the same way as #416


Darkren:
Maybe we should put this TestMain to some separate file with a conventional name? Now it's placed in some the test files for a package, which makes it hard to find in my opinion

PR #437, opened 2019-06-25
Fix transport not closing on DMSG server disconnect

### LINKED ISSUE #431: Transport is not being closed on DMSG server disconnect
Discussed as a part of DMSG server testing

Darkren:
Fixes #431


evanlinjin:
Although this fixes the issue at hand, but I will have to reject these changes because it duplicates instead of reuse code. I will post a PR soon to demonstrate a more elegant solution (in my opinion).

Please don't take offense in this; good work with figuring out the source of the problem.

PR #440, opened 2019-06-26
Fix lingering transports after dmsg.Server disconnects.

### LINKED ISSUE #431: Transport is not being closed on DMSG server disconnect

evanlinjin:
Re-implementation of #437 
Fixes #431


pkg/dmsg/client.go
```diff
@@ -258,21 +256,37 @@ func (c *ClientConn) DialTransport(ctx context.Context, clientPK cipher.PubKey)
...
-		_ = c.Conn.Close() //nolint:errcheck
 		c.mx.Unlock()
-		c.wg.Wait()
 	})
+	return closed
+}
+
+// Close closes the connection to dms_server.
+func (c *ClientConn) Close() error {
+	if c.close(false) {
```
Darkren: What if a server disconnects, client waits a bit, decides that a server is dead and calls `Close`? This way the `CLOSE` frame will be sent, since `false` is hardcoded

evanlinjin: @Darkren that way `CLOSE` won't be written and will return an error (which is ignored). Writing on a closed connection is generally fine, it just returns an error (from my understanding).

evanlinjin: I guess the point of having the boolean input is just to save some unnecessary operations and technically we can have it without.

Darkren: Yeah I thought of it. If we decide to live with writing to the closed connection, maybe it would be better just to leave one transports loop and call exported `Close` on each iteration

evanlinjin: Let's do that. I'll make changes.

Darkren (approved):
Looks good to me

pkg/dmsg/server_test.go
```diff
@@ -1127,6 +1127,7 @@ func testWithTimeout(timeout time.Duration, run func() error) error {
 			case <-timer.C:
 				return err
 			default:
+				time.Sleep(time.Millisecond * 5)
```
ayuryshev, line 1130: I would do something like
```golang
errCh := make(chan error)
go func(){
 errCh <- run()
}()

select{
case <- time.After(timeout):
  t.Fatal("timeout")
case <- errCh:
  return
}
```

evanlinjin: @ayuryshev I don't see how the proposed approach improves on what is already written. Hence, I'll resolve this conversation.

pkg/dmsg/client.go
```diff
@@ -271,8 +269,15 @@ func (c *ClientConn) Close() error {
 		}
```
ayuryshev, line 269: Comment for `if tp !=nil`: 
I see in standard lib a lot of times such checks inside `Close`-methods.
I would consider do the same way.

ayuryshev (approved):

pkg/dmsg/client.go
evanlinjin: Good point, I'll create a separate issue.

PR #450, opened 2019-06-27
Fix rare 'panic: send on closed channel' with dmsg.Transport

### LINKED ISSUE #449: `panic: send on closed channel` happens in DMSG transport
Consider an ongoing communication between DMSG clients. Suddenly server disconnects which make clients close their transports. Inside `Transport`'s `Serve` there is a frame type `switch` with the case responsible for handling `FWD` frames. Inside there is a piece of code:
```
				// notify of new data via 'bufCh'
				select {
				case tp.bufCh <- struct{}{}:
				default:
				}
```
When transport gets closed, it's `bufCh` also gets closed. If it happens almost simultaneously with this select (probably `case` should be triggered just before the channel gets closed) it leads to this panic. Extremely rare case but should be dealed with, since the panic may ruin experience with DMSG

evanlinjin:
Closes #449 

### Explanation of fix
The follow logic is implemented:
1. `done` is always closed before `inCh`/`bufCh` is closed.
2. mutexes protect `inCh`/`bufCh` to ensure that closing and writing to these chans does not happen concurrently.
3. Our worry now, is writing to `inCh`/`bufCh` AFTER they have been closed.
4. But as, under the mutexes protecting `inCh`/`bufCh`, checking `done` comes first, and we know that `done` is closed before `inCh`/`bufCh`, we can gurantee that it avoids writing to closed chan.


ivcosla (approved):
Looking good from my side

pkg/dmsg/transport.go
```diff
@@ -130,17 +130,17 @@ func (tp *Transport) Type() string {
...
+		return io.ErrClosedPipe
+	}
 	select {
 	case tp.inCh <- f:
 		return nil
 	default:
-		if tp.IsClosed() {
-			return io.ErrClosedPipe
-		}
 		goto handleFrame
```
Darkren, line 144: This way, we can now use `for` loop here instead of `goto`, would be more beatiful imo

Darkren (commented):
I think this should work, great job! Also, it may be a good idea to describe this logic in the comments, it's tricky

pkg/dmsg/transport.go
evanlinjin: @Darkren good points created here. As `dmsg` has moved, I will discuss this there!
https://github.com/skycoin/dmsg

PR #455, opened 2019-06-30
Fix data race in SSH

### LINKED ISSUE #435: SSH data race
Sometimes when executing ssh integration environment, and running ssh test a data race appears on the ssh server node. Attached are the logs for every node, the data race appears in A log, line 126.

[nodeA.log](https://github.com/skycoin/skywire/files/3325443/nodeA.log)
[nodeB.log](https://github.com/skycoin/skywire/files/3325444/nodeB.log)
[nodeC.log](https://github.com/skycoin/skywire/files/3325445/nodeC.log)

Darkren:
Fixes #435 . 

Implementation follows the same pattern as in the DMSG transport. Also added `sync.Once` to prevent duplicate `close` calls on channels if the `SSHChannel`'s `Close` gets called more than once accindetally.

### DIFF
--- a/internal/therealssh/channel.go
+++ b/internal/therealssh/channel.go
@@ -40,7 +40,8 @@ type SSHChannel struct {
 	dataChMx sync.Mutex
 	dataCh   chan []byte
 
-	done chan struct{}
+	doneOnce sync.Once
+	done     chan struct{}
 }
 
 // OpenChannel constructs new SSHChannel with empty Session.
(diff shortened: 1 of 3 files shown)


internal/therealssh/channel.go
```diff
@@ -35,12 +36,18 @@ type SSHChannel struct {
...
 	dataCh   chan []byte
+
+	doneOnce sync.Once
+	done     chan struct{}
 }
 
 // OpenChannel constructs new SSHChannel with empty Session.
 func OpenChannel(remoteID uint32, remoteAddr *app.Addr, conn net.Conn) *SSHChannel {
-	return &SSHChannel{RemoteID: remoteID, conn: conn, RemoteAddr: remoteAddr, msgCh: make(chan []byte), dataCh: make(chan []byte)}
+	return &SSHChannel{RemoteID: remoteID, conn: conn, RemoteAddr: remoteAddr, msgCh: make(chan []byte),
```
nkryuchkov, line 49: I think that writing every struct field on a new line might be more readable. What are your thoughts about it?

evanlinjin: I think in this case, it is fine. I would leave it as is ☺️

evanlinjin (approved):

PR #465, opened 2019-07-02
Fixed some routing logic.

evanlinjin:
* Fixed various race conditions in 'app' module.
* Fix chance where app.appConn does not get closed.
* Fixed various memory leaks in managedTransport (where channels are not closed).
* Improved transport logging so that we don't need to write to disk on every network operation.

### DIFF
--- a/integration/test-messaging.sh
+++ b/integration/test-messaging.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 source ./integration/generic/env-vars.sh
-# curl --data  {'"recipient":"'$PK_A'", "message":"Hello Joe!"}' -X POST  $CHAT_C  
+curl --data  {'"recipient":"'$PK_A'", "message":"Hello Joe!"}' -X POST  $CHAT_C
 curl --data  {'"recipient":"'$PK_C'", "message":"Hello Mike!"}' -X POST  $CHAT_A
--- a/integration/test-messaging.sh
+++ b/integration/test-messaging.sh
@@ -1,4 +1,3 @@
 #!/usr/bin/env bash
-source ./integration/generic/env-vars.sh
 curl --data  {'"recipient":"'$PK_A'", "message":"Hello Joe!"}' -X POST  $CHAT_C
 curl --data  {'"recipient":"'$PK_C'", "message":"Hello Mike!"}' -X POST  $CHAT_A
--- a/pkg/transport/log.go
+++ b/pkg/transport/log.go
@@ -2,6 +2,7 @@ package transport
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/big"
 	"os"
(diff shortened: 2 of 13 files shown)


integration/test-messaging.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 source ./integration/generic/env-vars.sh
```
ayuryshev: This is an error - env-vars.sh always from `./integration/generic`
And in case of PK/SK change they became incorrect.
It's better to comment or delete this line.

pkg/transport/log.go
```diff
@@ -14,8 +15,69 @@ import (
 // LogEntry represents a logging entry for a given Transport.
 // The entry is updated every time a packet is received or sent.
 type LogEntry struct {
-	ReceivedBytes *big.Int `json:"received"` // Total received bytes.
-	SentBytes     *big.Int `json:"sent"`     // Total sent bytes.
+	RecvBytes big.Int `json:"recv"` // Total received bytes.
```
ayuryshev: `RecvBytes, SentBytes uint64` - is not enough?
We can simplify this structure in radical if we use plain uint64 and atomic.AddUint64

pkg/transport/log.go
```diff
@@ -32,14 +94,17 @@ type inMemoryTransportLogStore struct {
 // InMemoryTransportLogStore implements in-memory TransportLogStore.
 func InMemoryTransportLogStore() LogStore {
 	return &inMemoryTransportLogStore{
-		entries: map[uuid.UUID]*LogEntry{},
+		entries: make(map[uuid.UUID]*LogEntry),
```
ayuryshev, line 86: If it would be possible to simplify LogEntry - this map could be simplified too into `map[uuid.UUID]LogEntry`

pkg/transport/manager.go
```diff
@@ -370,47 +370,68 @@ func (tm *Manager) isClosing() bool {
 	}
 }
 
-func (tm *Manager) manageTransport(ctx context.Context, mTr *ManagedTransport, factory Factory, remote cipher.PubKey, public bool, accepted bool) {
+func (tm *Manager) manageTransport(ctx context.Context, mTr *ManagedTransport, factory Factory, remote cipher.PubKey) {
```
ayuryshev, line 373: Good. I thought about similar change

pkg/transport/manager_test.go
```diff
@@ -218,25 +218,25 @@ func TestTransportManagerLogs(t *testing.T) {
...
+	tr1 := m1.Transport(tr2.Entry.ID)
 	require.NotNil(t, tr1)
 
 	go tr1.Write([]byte("foo")) // nolint
 	buf := make([]byte, 3)
 	_, err = tr2.Read(buf)
 	require.NoError(t, err)
 
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(time.Second * 10)
```
ayuryshev: 10 seconds - is not too much here?

pkg/transport/managed_transport.go
```diff
@@ -1,102 +1,106 @@
...
+	"time"
 )
 
 // ManagedTransport is a wrapper transport. It stores status and ID of
 // the Transport and can notify about network errors.
 type ManagedTransport struct {
 	Transport
-	ID       uuid.UUID
-	Public   bool
+	Entry    Entry
```
ayuryshev, line 15: This embedded `Entry Entry` - is what I like most in this PR.

ayuryshev (approved):
This PR looks good I like embeddeded `Entry Entry` in `ManagedTransport` the most.

pkg/transport/log.go
evanlinjin: You are right. I just did calculations. It will take 6881 years to max uint64 with 85mb/s (taiwan internet speed) and constant data. And I didn’t take into account metadata.

pkg/transport/manager_test.go
evanlinjin: This is to take into account how long it waits before we write the transport logs to disk.

pkg/transport/managed_transport.go
```diff
@@ -1,102 +1,106 @@
...
+		Transport: tr,
+		Entry:     entry,
+		Accepted:  accepted,
+		done:      make(chan struct{}),
+		update:    make(chan error, 16),
+		LogEntry:  new(LogEntry),
 	}
 }
 
 // Read reads using underlying
```
Darkren: I understand that it's an old piece, but the comment seems unfinished

pkg/transport/managed_transport.go
```diff
@@ -1,102 +1,106 @@
...
+		select {
+		case tr.update <- err:
+		default:
+		}
+	}
+	tr.mu.RUnlock()
 	return
 }
 
 // Write writes to an underlying
```
Darkren: Same as above

pkg/transport/managed_transport.go
```diff
@@ -1,102 +1,106 @@
...
 }
 
+func (tr *ManagedTransport) killUpdate() {
+	tr.mu.Lock()
+	close(tr.update)
+	tr.update = nil
+	tr.mu.Unlock()
+}
+
 // Close closes underlying
```
Darkren: Unfinished comment

Darkren (commented):
Besides what @ayuryshev mentiond, I'd finish some comments. Otherwise, great job!

pkg/transport/log.go
```diff
@@ -32,14 +94,17 @@ type inMemoryTransportLogStore struct {
...
 	}
 }
 
 func (tls *inMemoryTransportLogStore) Entry(id uuid.UUID) (*LogEntry, error) {
 	tls.mu.Lock()
-	entry := tls.entries[id]
+	entry, ok := tls.entries[id]
 	tls.mu.Unlock()
+	if !ok {
+		return entry, errors.New("not found")
```
nkryuchkov: Would you consider extracting the error to a package-level variable? This could let package users compare a returned error to the variable.

nkryuchkov (approved):
Looks awesome, great job!

pkg/transport/manager_test.go
ivcosla: I think should be fine

ivcosla (approved):
Good job!. However it needs format to pass the linter.

pkg/transport/log.go
evanlinjin: From my understanding, if we are doing concurrent write operations, we shouldn't copy the structure. I think it may be best if this is kept a pointer?

pkg/transport/log.go
evanlinjin: I think generally I would agree on this, but I think there will not be a use case for this scenario.

pkg/transport/log.go
ayuryshev: In general values are more safe for concurrency:
- value is copied and no possibility that innards of structure will be changed somewhere else

But `LogEntry` has `rMx,rWx       sync.Mutex` - they could not be copied and moved.

So I was wrong here: no choice exists  - only by pointer.

ayuryshev: But I see that LogEntry lost their mutexes.
Now `map[uuid.UUID]LogEntry` looks preferable.

evanlinjin: It's not just with mutexes. There is a possibility that we are in the middle of a `atomic.WHATEVER` operation and the structure get's copied because we need to write it to disk. From my understanding, that will lead to data race.

ayuryshev: Let's look at situation:

```golang
func funcA(){
  entry = &LogEntry{}
  funcB(entry)
  entry.RecievedBytes = 2000
}
func funcB(entry *LogEntry){
   entry.RecievedBytes = 1000
}
```
In this case we have possibility of race.

But if we change semantic of `funcB`  to `funcB(entry LogEntry)` - there will be no possibility of race. 
Because value will be copied and each function will have a different `entry`

PR #466, opened 2019-07-03
[WIP] [ROUTING] Remove 'transport.Manager' dependency from 'setup.Node'. 

### LINKED ISSUE #442: [ROUTING] Remove 'transport.Manager' dependency from 'setup.Node'.
Setup Node should not be using `transport.Manager`, and should use `dmsg.Client` directly.

**Tasks:**
- [x] Remove `transport.Manager` from `setup.Node` completely.
- [x] Come up with plan as to how this is to be tested.
- [x] Refactor current tests to be more understandable (or remove them).

nkryuchkov:
Closes #442


pkg/setup/node.go
```diff
@@ -84,22 +80,23 @@ func (sn *Node) Serve(ctx context.Context) error {
...
-			if tr.Accepted {
-				go func(t transport.Transport) {
-					for {
-						if err := sn.serveTransport(t); err != nil {
-							sn.Logger.Warnf("Failed to serve Transport: %s", err)
-							return
-						}
-					}
-				}(tr)
+		for {
```
evanlinjin: This for loop should not be in a go routine. As previously, we had `tm.Serve()` that blocks, we need to keep the same behavior.

pkg/transport/manager.go
```diff
@@ -265,10 +282,15 @@ func (tm *Manager) dialTransport(ctx context.Context, factory Factory, remote ci
 		return nil, nil, err
 	}
 
-	entry, err := settlementInitiatorHandshake(public).Do(tm, tr, time.Minute)
-	if err != nil {
-		tr.Close()
-		return nil, nil, err
+	var entry *Entry
+	if tm.IsSetupTransport(tr) {
```
evanlinjin, line 357: We also need a check when we accept remote transports.

evanlinjin: To elaborate, we need to check whether the accepted transport is setup or not. And if the transport is of setup, then don't do settlement handshake.

nkryuchkov: @evanlinjin Thanks for having a look. I've added a check to `acceptTransport`. Could you please look if this is a correct approach?

evanlinjin: It looks fine at a glance. Is the system working now?

nkryuchkov: @evanlinjin No. There're some errors when running integration tests. I'm trying to fix them.

evanlinjin:
@nkryuchkov From my testing, the system is working with `generic environment` and `./integration/test-messaging-loop.sh` script. Good job!

However, we have the issue of `visor-node`s constantly trying to retain transports with `setup-node`. I'm assuming this is because we are still using `managedTransport` with setup transports.

We probably need a separate field in `transport.Manager`: `SetupTpCh chan Transport` and rename `TrChan` to `DataTpCh`.

Here are some attached screenshots of my testing:
image
image
image
image

nkryuchkov:
@evanlinjin Thank you for your help! I get the same error messages. I'll try the approach you suggested.

pkg/transport/manager.go
```diff
@@ -364,9 +403,11 @@ func (tm *Manager) acceptTransport(ctx context.Context, factory Factory) (*Manag
 	select {
 	case <-tm.doneChan:
 		return nil, io.ErrClosedPipe
-	case tm.TrChan <- mTr:
+	case dataTpChan <- mTr:
```
evanlinjin: We probably need an if/else statement here, otherwise it is like playing duck-duck-goose. If it is a tp with purpose of setup, push it to `setupTpCh`, else, push it to `dataTpCh`.

pkg/transport/manager.go
```diff
@@ -333,14 +362,24 @@ func (tm *Manager) acceptTransport(ctx context.Context, factory Factory) (*Manag
...
-	entry, err := settlementResponderHandshake().Do(tm, tr, 30*time.Second)
-	if err != nil {
-		go func() {
-			if err = tr.Close(); err != nil {
-				tm.Logger.Warnf("Failed to close transport: %s", err)
-			}
-		}()
-		return nil, err
+	var setupTpChan chan Transport
+	var dataTpChan chan *ManagedTransport
```
evanlinjin: What are lines 365/366 for?

nkryuchkov: @evanlinjin If the transport is `setup`, the setup channel is enabled. If `data`, then the data channel. Since sending to a `nil` channel blocks forever, the `case` with a `nil` channel in `select` is not executed. But probably it may be rewritten in a more straightforward way with `if`/`else` as you suggested here: https://github.com/skycoin/skywire/pull/466#discussion_r304949768

evanlinjin:
```
 Since sending to a nil channel blocks forever, the case with a nil channel in select is not executed.
```
From my understanding, this is wrong. `select` statement is random, and if we are unlucky and it chooses the case with the `nil` chan, then it will, as you have stated, **block forever**. This is not the wanted behavior.

nkryuchkov: @evanlinjin As far as I know `select` chooses a random case not from all cases, but from ones that are ready for execution (https://golang.org/ref/spec#Select_statements), but a `nil` channel never is.

But anyway, I'm going to rewrite it because it looks complicated. Thanks for pointing out :)

pkg/transport/manager.go
nkryuchkov: @evanlinjin Thanks. Fixed.

evanlinjin:
I'm going to put this PR on hold for the time being. Really depends if these changes gets approved or not: https://github.com/skycoin/skywire/issues/488

evanlinjin:
Replaced by #494

PR #467, opened 2019-07-03
Remove encryption from routes

### LINKED ISSUE #463: [ROUTING] Remove encryption from routes.
Looking at how Brandon intends on how routes should function, doing encryption in the manner that we currently are, does not make sense.

Removing encryption from routes will also simplify the codebase.

Darkren:
closes #463 

Removed noise message and response from the funcs connected with a loop establishment. Therefore, removed handshake going on during the route establihsment


evanlinjin (approved):
Good work.

nkryuchkov (approved):
Looks good, great job!

ayuryshev (approved):

PR #472, opened 2019-07-04
added dockerfiles

ivcosla:
Dockerfiles for skywire nodes


ivcosla:
WIP until I figure out what is wrong with apps, which the node cannot find

docker/images/builder/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ARG CGO_ENABLED=0
+ENV CGO_ENABLED=${CGO_ENABLED} \
+    GOOS=linux  \
+    GOARCH=amd64 \
+    GO111MODULE=on
+
+COPY . skywire 
+WORKDIR skywire
+
+RUN ./docker/images/builder/update.sh
```
ayuryshev: Those 4 RUNs create 3 layers in image.
They could be aggregated into 
```
RUN [instruction_1] && [instruction_2] ... 
```

docker/images/builder/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN ./docker/images/builder/install.sh
```
ayuryshev: There is no need in skywire to have separate update.sh and install.sh.
install.sh already has everything from update.sh.

docker/images/builder/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN apk add make
```
ayuryshev: Separate RUN just to install make. Note that it will work only if apk  present

docker/images/builder/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+RUN go build -mod=vendor -tags netgo -ldflags="-w -s" \
```
ayuryshev: I would rather move it into compile.sh/build.sh - same as in skywire-services.

docker/images/node/Dockerfile
```diff
@@ -0,0 +1,12 @@
+ARG builder=node-builder
+ARG base=alpine
+
+FROM ${builder} as builder
```
ayuryshev: builder/Dockerfile could be merged with this Dockerfile.

Merge and add last line:
```
FROM base
```

docker/images/node/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+FROM ${base} as base
+
+COPY --from=builder /go/skywire/skywire-node skywire-node
+COPY --from=builder /go/skywire/apps bin/apps
+COPY --from=builder /go/skywire/docker/images/builder/update.sh update.sh
```
ayuryshev: Why update.sh needed in built image?

ayuryshev (changes requested):
Good. Some cleanup needed.

docker/images/node/update.sh
```diff
@@ -0,0 +1,16 @@
...
+        && rm -rf /var/lib/apt/lists/* 
+fi
+
+if type apk > /dev/null; then
+        
+        apk update 
+        apk upgrade 
+        apk add --no-cache ca-certificates openssl iproute2
+        update-ca-certificates --fresh
+        apk add iproute2
```
ayuryshev, line 15: `iproute2` second time here

ayuryshev (commented):
Good

ayuryshev (approved):

PR #480, opened 2019-07-12
Format JSON files

nkryuchkov:
JSON files had different code style. Formatted by `Reformat code` option with default settings in Goland. 2 spaces indentation.


ivcosla (approved):
Good catch

PR #481, opened 2019-07-14
Remove linter excluding rules, clean up code

### LINKED ISSUE #460: Include more errors into lint check
I see that we have `- Error return value of .((os\.)?std(out|err)\..*|.*Close|.*Flush|os\.Remove(All)?|.*printf?|os\.(Un)?Setenv). is not checked` in the linter config. 

Although I understand that some of the errors are unlikely to ever happen, I suggest that we remove this or at least include some of its parts and modify the code to be more robust in case of such unexpected scenarios. Such an approach leads to heavy error ignoring all around the code (my concern arised when I took a look at `therealssh` code)

nkryuchkov:
Closes #460.

Changes:
- Removed all linter excluding rules
- Removed all `nolint` comments except a couple of them that are really necessary
- Fixed all linter issues after removal
- Made some code cleanup
- Fixed issues found by `gocritic`, `scopelint`, `unconvert` linters


internal/therealproxy/server_test.go
```diff
@@ -36,7 +35,7 @@ func TestProxy(t *testing.T) {
 	srv, err := NewServer("")
 	require.NoError(t, err)
 
-	l, err := net.Listen("tcp", ":10081") // nolint: gosec
+	l, err := net.Listen("tcp", "localhost:10081")
```
Darkren: I'd suggest this to be implemented via `nettest`

pkg/httputil/httputil.go
```diff
@@ -8,15 +8,18 @@ import (
...
+var log = logging.MustGetLogger("httputil")
+
 // WriteJSON writes a json object on a http.ResponseWriter with the given code,
 // panics on marshaling error
 func WriteJSON(w http.ResponseWriter, r *http.Request, code int, v interface{}) {
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(code)
 	enc := json.NewEncoder(w)
-	if pretty, _ := BoolFromQuery(r, "pretty", false); pretty { //nolint:errcheck
+	if pretty, err := BoolFromQuery(r, "pretty", false); err == nil && pretty {
```
Darkren: Shouldn't this error be logged?

pkg/setup/node_test.go
```diff
@@ -285,18 +301,23 @@ func (f *muxFactory) Type() string {
 
 type mockNode struct {
 	sync.Mutex
-	rules map[routing.RouteID]routing.Rule
-	tm    *transport.Manager
+	testing *testing.T
```
Darkren: It may ruin tests if we create a single `mockNode` and run several `t.Run` cases against it

Darkren (changes requested):
Overall it's a huge job done and done great, but I left some comments

pkg/setup/node_test.go
nkryuchkov: @Darkren well spotted, I'm going to get rid of `testing *testing.T` field

pkg/httputil/httputil.go
nkryuchkov: @Darkren Probably a good idea, thanks. Added some logging.

internal/therealproxy/server_test.go
nkryuchkov: @Darkren Good idea. Done.

cmd/apps/therealproxy-client/therealproxy-client.go
```diff
@@ -30,7 +30,11 @@ func main() {
 	if err != nil {
 		log.Fatal("Setup failure: ", err)
 	}
-	defer socksApp.Close()
+	defer func() {
+		if err := socksApp.Close(); err != nil {
+			log.Println("Failed to close app: ", err)
```
evanlinjin: No need for space after `app: `. It is `Println` after all.

internal/httpauth/client.go
```diff
@@ -126,10 +133,16 @@ func (c *Client) Nonce(ctx context.Context, key cipher.PubKey) (Nonce, error) {
 	req = req.WithContext(ctx)
 
 	resp, err := c.client.Do(req)
+	if resp != nil {
```
evanlinjin, line 136: Shouldn't we check the `err` first?

internal/httpauth/client_test.go
```diff
@@ -55,8 +54,10 @@ func TestClient(t *testing.T) {
 	require.NoError(t, err)
 
 	b, err := ioutil.ReadAll(res.Body)
+	if b != nil {
```
evanlinjin: Shouldn't we be checking the error here first?

internal/httpauth/client_test.go
```diff
@@ -83,8 +84,10 @@ func TestClient_BadNonce(t *testing.T) {
...
```
evanlinjin: Check error first.

internal/httpauth/client.go
nkryuchkov: @evanlinjin Both `resp` and `err` may be non-nil in case of a redirection failure.
Source: http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#close_http_resp_body

cmd/apps/therealproxy-client/therealproxy-client.go
nkryuchkov: @evanlinjin Well spotted, thanks.

internal/httpauth/client_test.go
nkryuchkov: @evanlinjin Actually, this looks strange, I'm going to rewrite it.

evanlinjin (approved):

PR #482, opened 2019-07-15
Routing data type cleanup

### LINKED ISSUE #444: [ROUTING] Data type cleanup.
We are using similar data types that can be combined to simplify procedures. The following is not a fully formulated list and someone needs to dive into the code to see what we can combine.

From my understanding:
- We were hiding the local node's public key from the apps (this was a pointless decision). Instead, we can just have a unified `Loop` structure that is used at all levels of the protocol.

**TODO: Finish formulating this ticket.**

Darkren:
Fixes #444 

- `app.Addr` and `app.LoopAddr` were combined and moved to `routing.Addr`
- type `Port uint16` was introduced
- `setup.LoopData` was refactored to use `routing.Addr` and moved to `routing.LoopData`
- `routing.Loop` and `routing.LoopDescriptor` are separate type now


nkryuchkov (approved):
Looks good to me, nice job! However, I haven't tested it.

internal/therealssh/client.go
```diff
@@ -15,14 +15,14 @@ import (
...
 
 	"github.com/skycoin/skywire/internal/netutil"
-	"github.com/skycoin/skywire/pkg/app"
+	"github.com/skycoin/skywire/pkg/routing"
 )
 
 var r = netutil.NewRetrier(50*time.Millisecond, 5, 2)
 
 // Dialer dials to a remote node.
 type Dialer interface {
```
ayuryshev: I have questions about Dialer interface:

I see 2 types that implements Dialer interface: 
- `therealsssh.pipeDialer`  - used in tests only
- `app.App` - in application code

`Dialer` as argument is used only in `therelealssh.NewClient`

Question 1: Why this interface made public? Is it intended to use it outside `therealssh` ?
Question 2: In case  `Dialer` is intended to be used outside: maybe it's makes sense to move it in `app` package?

internal/therealssh/client.go
```diff
@@ -15,14 +15,14 @@ import (
...
-	Dial(raddr *app.Addr) (net.Conn, error)
+	Dial(raddr *routing.Addr) (net.Conn, error)
```
ayuryshev: Why not define without pointer as `Dial(raddr routing.Addr) (net.Conn, error)` ? 

I see that (almost) everywhere that `raddr` is constructed on place in a manner: 
`Dial(&routing.Addr{pk, port})`.  

It's a possible place for data races

internal/therealssh/client.go
```diff
@@ -54,7 +54,7 @@ func (c *Client) OpenChannel(remotePK cipher.PubKey) (localID uint32, sshCh *SSH
 	var err error
 
 	err = r.Do(func() error {
-		conn, err = c.dialer.Dial(&app.Addr{PubKey: remotePK, Port: Port})
+		conn, err = c.dialer.Dial(&routing.Addr{PubKey: remotePK, Port: Port})
```
ayuryshev: And here is an example of `*routing.Addr` construction on place.
Is it possible to change signature of `Dial` to `func(routing.Addr) (net.Conn,error)` ?

pkg/routing/addr.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// Port is a network port number
+type Port uint16
+
+const networkType = "skywire"
+
+// Addr represents a network address combining public key and port.
+// Implements net.Addr
+type Addr struct {
```
ayuryshev, line 16: Question about lifecyle for values of `Addr` type:
- Is this type intended to be used as immutable: created once and then never changed internals?
- Or intentions is to use as:
```golang
adrr:= Addr{Port(13)}
go somefunc(*addr)
addr.Port = Port{14}
```
?

So far I found only one place of possible mutation: `app.Protocol.Send` where pointer to Addr passed as `interface{}`

Question is not about `Addr` type itself but about most methods that use `*routing.Addr` params

pkg/routing/loop.go
```diff
@@ -7,17 +7,27 @@ import (
 	"github.com/skycoin/dmsg/cipher"
 )
 
-// Loop defines a loop over a pair of routes.
+// Loop defines a loop over a pair of addresses.
 type Loop struct {
```
ayuryshev, line 11: The same question as for `Addr` arises for `Loop` and `LoopDescriptor`:
Are those types intended as mostly immutable or it's intended usage that internal content could be changed during lifetime?

In case it intended as mostly immutable - why not use them in functions as values without pointers?

ayuryshev (changes requested):
I suppose that this cleanup could be improved further with another step - changing signatures of  functions that use types `*routing.Addr`, `(routing.Loop` and `*routing.LoopDescriptor` for params to signatures without pointers.

E.g.
```golang 
type Dialer interface{
   Dial(raddr routing.Addr) (net.Conn, error)
   // instead of Dial(raddr *routing.Addr) (net.Conn, error)
}
```
This way it will automatically exclude posibilities for data races.

internal/therealssh/client.go
evanlinjin: `Dialer` is not intended to be used publicly.

internal/therealssh/client.go
evanlinjin: Yes it will be better to be a non-pointer.

internal/therealssh/client.go
evanlinjin: Yes.

pkg/routing/addr.go
evanlinjin: It is better to be used as a non-pointer. Thanks for pointing this out. @Darkren would it be too much of a hassle to make the relevant changes?

pkg/routing/loop.go
evanlinjin: Immutable.

evanlinjin (commented):
Looks mostly good. Would merge after the changing the associated structures to be immutable (as mentioned by @ayuryshev ).

pkg/routing/addr.go
Darkren: @evanlinjin I think from this point of view this should be done regardless of time. I don't think this will take too much anyway, I'll do it

internal/therealssh/client.go
ayuryshev: If that's the case - why not `type dialer interface`?

Darkren: I also moved a bit further, made `routing.LoopDescriptor` and `routing.LoopData` being passed as non-pointers too. I believe that this is appropriate too. So, the only thing left is this question about `dialer` @evanlinjin

evanlinjin: @Darkren let's make `Dialer` private as @ayuryshev suggested.

Darkren: @evanlinjin made it unexported, also moved to a separate `dialer.go` file

evanlinjin (approved):
Everything looks good and seems to be working fine according to my tests.

PR #483, opened 2019-07-15
Improve SSH logging

### LINKED ISSUE #456: Improve SSH logging
Notices this piece of code in `therealssh`:
```
func debug(format string, v ...interface{}) {
	if !Debug {
		return
	}

	log.Printf(format, v...)
}
```
And it's actually being heavily used. I suggest that we improve logging here and switch to more common solutions which support log levels like it's done in the other parts (i.e. DMSG)

Darkren:
Fixes #456 

Removed `debug` func, used the `*logging.Logger` instead. If the `Debug` flag is not being passed, thus making `Debug` equal false, then the logging level is being set to info, to omit all the debug output. Also made general changes to the logging


cmd/apps/therealssh/therealssh.go
```diff
@@ -32,6 +34,9 @@ func main() {
 	}
 
 	ssh.Debug = *debug
+	if !ssh.Debug {
+		logging.SetLevel(logrus.InfoLevel)
```
nkryuchkov, line 42: Would you consider getting rid of a `logrus` dependency here using `logging.LevelFromString`? This way abstracts the logger implementation from this package. I'd write the following:
```suggestion
		if infoLevel, err := logging.LevelFromString("info"); err == nil {
			logging.SetLevel(infoLevel)
		}
```

nkryuchkov (approved):
Looks good to me, well done! I'd suggest a small change, please decide if it's needed.

cmd/apps/therealssh/therealssh.go
evanlinjin: @nkryuchkov I personally think this is not necessary. `logging` depends on `logrus` anyway so we won't we having unnecessary dependencies.

evanlinjin (approved):
Looks good to me.

evanlinjin:
@Darkren can you please resolve the conflicts before we merge? Thanks!

evanlinjin (approved):

PR #494, opened 2019-07-20
[ROUTING] Remove transport manager dependency.

### LINKED ISSUE #442: [ROUTING] Remove 'transport.Manager' dependency from 'setup.Node'.

evanlinjin:
closes #442


ivcosla (approved):
Good job!

nkryuchkov (approved):
Good job! Looks good to me.

ivcosla (commented):
I integration-tested it.

In messaging integration environment and ssh integration environment:

Works faster than before. Quite faster. On ssh environment I have not experienced any error, restarting every kind of node it's always able to ssh from C to A again.

However, in messaging environment things break when either A or C are restarted (after having succesfully sent one message between them). The first message send after that appears in the app as delivered but doesn't makes it to the other end. The second message displays an `read/write error on closed Pipe`. Manually re-adding the transports between the nodes fixes the problems.

ayuryshev:
# What I did

1. In dmsg repo: `git checkout evanlinjin/feature/improve-transport-logic`
2. In skywire repo 
```sh
$ git checkout evanlinjin/feature/remove-transport-manager-dependency
$ make mod_uncomm ## checked go.mod that  dmsg is replaced
$ make dep && make build && make test
```

# What I saw

1. Both repos are in detached state. They need to be remerged: skywire with mainnet and dmsg with master
2. I looked again for travis-ci test and I see that only license was checked. Travis did not start test with branch with conflicts
3. In results of make test I see 
```
--- FAIL: ExampleManager_CreateDataTransport (2.22s)
```
But looking for output I see that FAIL must possibly not from the test itself but from debugging `Printf`s that pollute output with messages like:
```
{tp.Manager} Factory found tp: local(032bce8baccfb05963de4f88b3648e4ead0237c547070deb6ca21c223b4bfa28fd) remote(023acb3d969fda6865bcb3295b931dc86914f655c9bf4adb053e6b422fbde632bf)
```
when test expects only  `Manager.CreateDataTransport success`
which is printed too.

So I suppose that tested functionality is working as expected  - it's the problem in test itself.

# Overall

1. Looks like close to be finished 

2. Needs to be remerged with base branches to be travis-tested 

3. ExampleManager_CreateDataTransport:
  - either remove  prints that pollute output
  - or convert from Examle to Test 
both variants will make it pass

PR #498, opened 2019-07-22
Run linters in vendor mode

### LINKED ISSUE #497: Linters do not use dependencies inside `vendor` folder
Vendoring mode is not enabled in the `golangci-lint` config and the `go vet` flags. This sometimes leads to broken `Travis CI` builds when dependencies in `vendor` folder and repository are not the same, because `golangci-lint` and `go vet` do not use the `vendor` folder.

An example of a broken build:
https://travis-ci.com/skycoin/skywire/builds/120084033
This happened in https://github.com/skycoin/skywire/pull/496 on commit https://github.com/skycoin/skywire/pull/496/commits/f477fb1a558e0630b5ed69afdaf63c9a0052e9b4. Here, linters cannot find the changed code in the `vendor` folder and try to download it instead which leads to a broken build.

nkryuchkov:
Closes #497


Makefile
```diff
@@ -45,7 +45,7 @@ rerun: stop
 lint: ## Run linters. Use make install-linters first	
 	${OPTS} golangci-lint run -c .golangci.yml ./...
 	# The govet version in golangci-lint is out of date and has spurious warnings, run it separately
-	${OPTS} go vet -all ./...
+	${OPTS} go vet -mod=vendor -all ./...
```
ayuryshev, line 48: Golang wiki recomends to state this flag explicitly. Although I did not saw misbeahavior without this flag - it's better to set it than not.

ayuryshev (approved):

Issue #501, opened 2019-07-25
Implement ability to switch between transport types (dmsg and tcp transport).


ayuryshev:
**Feature description**

Implement ability to specify between different underlying network connection types when saving transports.

**Important notes**

Communication between setup nodes and visor nodes **SHOULD ALWAYS** be via `dmsg.`
Electable network types should only be done for transports.

**Tasks**

These tasks should be done in the given order.

- [ ] Branch from [`mainnet-milestone1`](https://github.com/skycoin/skywire/tree/mainnet-milestone1).
- [ ] Move `TCPTransport` stuff to [`pkg/network`](https://github.com/skycoin/skywire/tree/mainnet-milestone1/pkg/network) module.
- [ ] Add needed fields to `network.Config`.
- [ ] Add `tcpC` member to `network.Network`.
- [ ] Add needed actions to `Network.Init()` (if needed).
- [ ] Add needed actions to `Network.Close()`.
- [ ] Add `Network.TCP() *TCPFactory` field.
- [ ] Modify `Network.Dial()` and `Network.Listen()` to be able to dial/listen with `"tcp"`.
- [ ] Add `"tcp"` options to:
  - https://github.com/skycoin/skywire/blob/b5909ebadb24f19f3f66a83f764fc2fa36b0e997/pkg/visor/visor.go#L134
  - https://github.com/skycoin/skywire/blob/b5909ebadb24f19f3f66a83f764fc2fa36b0e997/pkg/visor/visor.go#L154
- Add `public-key:ip-address` table to `visor.Config` which in turn gets copied to `network.Config` and then used by `network.TCPFactory`.
- [ ] Add interactive tests to use `TCPFactory`.

PR #503, opened 2019-07-25
[WIP] Feature: TCPFactory/TCPTransport instead of dmsg

ayuryshev:
It's WIP.
Just compiled and passes regular tests.

Closes: #501


pkg/setup/node.go
```diff
@@ -30,7 +30,7 @@ type Node struct {
 	Logger *logging.Logger
 
 	tm        *transport.Manager
-	messenger *dmsg.Client
+	messenger transport.Factory
```
jdknives: The setup node is supposed to use dmsg exclusively and directly. The communication between the visor and the setup node should not use TCP transport.

jdknives (changes requested):

pkg/router/router.go
```diff
@@ -322,18 +327,33 @@ func (r *Router) requestLoop(appConn *app.Protocol, raddr routing.Addr) (routing
...
+			}
+		}()
+
+		r.Logger.Infof("Router.requestLoop 6\n")
+		if err := setup.CreateLoop(proto, ld); err != nil {
+			return routing.Addr{}, fmt.Errorf("route setup: %s", err)
+		}
+	case "tcp-transport":
+		r.Logger.Info("Skipping setup for tcp-transport")
+		_, err := r.tm.CreateSetupTransport(context.Background(), raddr.PubKey, "tcp-transport")
```
evanlinjin: `dmsg` is to be used with `SetupNode` always.

evanlinjin:
Just a thought, let's split this PR into two.
- PR1: Implement the ability to choose between transport types for estabilishing data transports.
- PR2: Testing stuff.

Let's make this PR be `PR1` so we can have it merged faster.

PR #505, opened 2019-08-02
Feature/therealssh pty

### LINKED ISSUE #500: Therealssh-cli bypass TTY consuming
In the same way that unix `ssh` we need a way of make it work without having a TTY. This is in order to write integration tests with golang.

We need a similar solution as the flags `-t` and `-T` described in the following manual page: https://www.unix.com/man-page/redhat/1/ssh/.

ivcosla:
Closes #500


pkg/therealssh/pty_test.go
```diff
@@ -0,0 +1,80 @@
...
+			Y:    100,
+		},
+	}
+	_, err = ch.Request(RequestPTY, args.ToBinary())
+	require.NoError(t, err)
+
+	_, err = ch.Request(RequestExecWithoutShell, []byte("ls"))
+	require.NoError(t, err)
+
+	b := make([]byte, 6024)
```
nkryuchkov, line 53: Does `6024` mean anything or is it just an arbitrary value?

nkryuchkov (approved):
Looks good to me. Great job!

pkg/therealssh/pty_test.go
```diff
@@ -0,0 +1,80 @@
...
+	dialConn, acceptConn := net.Pipe()
+	pd := PipeDialer{PipeWithRoutingAddr{dialConn}, acceptConn}
+	_, client, err := NewClient(":9999", pd)
+	require.NoError(t, err)
+
+	server := NewServer(MockAuthorizer{})
+	go func() {
+		err := server.Serve(PipeWithRoutingAddr{acceptConn})
+		fmt.Println("server.Serve finished with err: ", err)
+		require.NoError(t, err)
```
Darkren: I think the error should be better passed outside of a goroutine and checked there. This one is probably won't work

Darkren (changes requested):
Otherwise, great job!

pkg/therealssh/pty_test.go
evanlinjin: Yes, we should never use `require` or `assert` library inside a go routine as it may result in a race condition.

pkg/therealssh/channel.go
```diff
@@ -230,6 +232,26 @@ func (sshCh *SSHChannel) Start(command string) error {
...
+		return errors.New("session is not started")
+	}
+
+	out, err := sshCh.session.Run(command)
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		_, err = sshCh.Write(out)
```
evanlinjin: Reusing `err` may result in race condition as we are reading `err` in the return statement later. I would add a colon:
```golang
_, err := sshCh.Write(out)
```

pkg/therealssh/session.go
```diff
@@ -77,6 +78,37 @@ func (s *Session) Start(command string) (err error) {
 	return cmd.Start()
 }
 
+// Run executes a command and returns it's output and error if any
+func (s *Session) Run(command string) ([]byte, error) {
+	var err error
```
evanlinjin, line 83: I don't think we need this.

evanlinjin (commented):
Good stuff! These are mode two cents for code quality. I'll actually test it out next.

pkg/therealssh/pty_test.go
ivcosla: Arbitrary value. Wanted to make sure it was big enough

pkg/therealssh/session.go
ivcosla: We need it to overwrite `command`. Otherwise in line 86 `command` is taken as a new variable if we use a colon in there.

pkg/therealssh/pty_test.go
ivcosla: `server.Serve` is not returning after neither `server.Close` or `dial/acceptConn.Close`, so for now I just turned off the linter.

ivcosla:
Commented this test https://github.com/skycoin/skywire/pull/505/files#diff-858b5f577c60d24dea6b5c413b8452b3R17 since it's messing with parallel execution. Only one rpc client with the same name can exist at a time, so when both tests are launched one fails.

If you think this test is not needed since it's logic is also tested in the next test let me know and if the PR is approved I will just delete it and merge.

pkg/therealssh/session.go
evanlinjin: Okay makes sense.

Issue #507, opened 2019-08-06
Transport fails to update while restarting visors


Darkren:
From discord:

"There is a bug with the current state of skywire. When running the messaging interactive test, and randomly restarting visor nodes, sometimes transports fail to get updated properly. I'm assuming it has to do with the logic of how transport statuses as submitted to TransportDiscovery."

Issue #443, opened 2019-06-26
[ROUTING] Decouple 'managedTransport' from 'transport.Manager'.


evanlinjin:
**Summary:**
- Rename `transport.managedTransport` to `transport.managedDataTp` (as we will need `transport.managedSetupTp` in the future).
- `transport.managedDataTp` should be self-sufficient, and shouldn't depend on `transport.Manager` for restarts.

**Motivation:**
- Have `transport.managedDataTp` be fully responsible for re-establishment.
- Have transport establishment/reestablishment function in a more predictable manner.

## Changes

**In regard to Transport IDs:**
- Transport IDs should be derived from local/remote public keys and the transport type ONLY. Private data transports are not to be registered.
- The least-significant edge should ALWAYS be the initiator.
- The least-significant edge is responsible for re-establishing the transport.
- Maybe using the `big` module to determine least-significant edge will be simpler.

**Regarding the Settlement Handshake:**
- `managedDataTp` should be responsible for performing the settlement handshake.
- Only data-transports are required to register themselves in `transport.Discovery`.

**Regarding Read/Write operations:**
- Any failure to a transport's read/write should close the associated route (in which the read/write failed).

**Regarding data-transport logging:**
- We also need to remove read/write log chans. Records of read/write should be to a stored `big.Int` value (protected by mutex), and saved to disc in given intervals. The motivation for this is that with the current implementation, the `readLogChan` and `writeLogChan` is not closed, leading to memory leak.

**Regarding `transport.Manager` changes:**
- We should have two maps of `managedDataTp`s: `map[uuid.UUID]*managedDataTp` and `map[cipher.Pubkey]*managedDataTp`.
- The aforementioned maps should be filled with saved transports on startup (even though we haven't actually connected).
- Transports that are not data-transports do not need to be stored (for now).

evanlinjin:
This ticket needs to be split into multiple tickets. I'm in the process of that now.

jdknives:
@evanlinjin this is fixed by #508 ? If so please close it, if not lets discuss, whether this is relevant for the next Milestone.

Issue #271, opened 2019-04-05
Improve transport-settlement handshake.


evanlinjin:
> Depends on PR:#252

## Description

We should depend on transport-discovery to settle transports, otherwise it may lead to some ambiguity. The following changes depend on deterministic Transport IDs (implemented within PR:#252).

The transport-settlement handshake **SHOULD** look as follows:

![IMG_2327](https://user-images.githubusercontent.com/14065705/55617508-d8907e00-57c6-11e9-9fd8-9189880b17b7.jpg)

> **Legend**
> `TpDisc` - Transport Discovery.
> `Init-er` - Transport Initiating Node.
> `Resp-er` - Transport Responder Node.
> `E` - Transport Entry.
> `S` - Signature.
> `TpID` - Transport ID.
> `Wait(<v>)` - Repetitively query until the resource is obtained.

The current implementation of the transport-settlement handshake works very similarly to the aforementioned. The only differences include:

- The `Resp-er` replies with `E+2S` (Entry with two signatures) instead of `TpID` (Transport-ID).
- The `Resp-er` does not wait for the `TpDisc` to have the entry submitted.

### Further Notes

For the implementation, ensure that:
- Only **NEWLY-CREATED** transports should go through the settlement-handshake. Always check the Transport Discovery whether the transport already has an entry. Both edges of the settlement-handshake should check this.
  - The *Initiator* should check with the Transport Discovery whether a settlement-handshake is needed.
  - The *Responder* should check the Transport Discovery whenever they `Factory.Accept()` a transport, if a settlement-handshake is required.

## Tasks
- [ ] Make the aforementioned changes.
- [ ] Update tests.
- [ ] Update Skywire Specifications.

jdknives:
@evanlinjin this should simplify the transport establishment. However, if the tpID is deterministic, then why would `Resp-er` even respond with the tpID? Would that not just increase the number of messages?

evanlinjin:
@jdknives 
1. We need a way for `Responder` to inform `Initiator` that it accepts the request.
2. Given the assumption that `Responder` will always agree to establish the transport, the alternate solution would be to have `Initiator` repetitively poll `tpDisc` as it awaits the confirmation of the `Entry` being submitted. Or have `Initiator` request `tpDisc` with a timeout, and the `tpDisc` keeps the connection with `Initiator` open until the `Entry` is submitted by `Responder`. I am not a fan of either of these solutions as it adds more logic and complexity than what I have suggested.

PR #508, opened 2019-08-07
[WIP] Decouple managedTransport and improve transport handling logic.

evanlinjin:
Related to #443, #507 and #271

## Changes
* Fix transport re-establishment logic.
* Moved much of the logic from `transport.Manager` to `managedTransport` to improve clarity.
* Settlement Handshake: only the responder updates the status to transport discovery.
* Removed the ability to register 'private' transport.
* Better use of `context.Context` throughout the codebase.
* Added timeouts to `setup` library to avoid SetupNode hanging.
* Changed various golang examples to proper tests.
* Cleaned up some logging.
* Cleaned up some tests.
* TODO: Some `router` tests are failing and have been commented out. These need to be fixed at a later stage.

### DIFF
--- a/pkg/router/route_manager.go
+++ b/pkg/router/route_manager.go
@@ -35,7 +35,9 @@ func (rm *routeManager) GetRule(routeID routing.RouteID) (routing.Rule, error) {
 		return nil, errors.New("unknown RouteID")
 	}
 
-	if len(rule) < 13 {
+	// TODO(evanlinjin): This is a workaround for ensuring the read-in rule is of the correct size.
+	// Sometimes it is not, causing a segfault later down the line.
+	if len(rule) < routing.RuleHeaderSize {
 		return nil, errors.New("corrupted rule")
 	}
 
--- a/pkg/router/route_manager.go
+++ b/pkg/router/route_manager.go
@@ -35,6 +35,10 @@ func (rm *routeManager) GetRule(routeID routing.RouteID) (routing.Rule, error) {
 		return nil, errors.New("unknown RouteID")
 	}
 
+	if len(rule) < 13 {
+		return nil, errors.New("corrupted rule")
+	}
+
 	if rule.Expiry().Before(time.Now()) {
 		return nil, errors.New("expired routing rule")
 	}
--- a/pkg/setup/node.go
+++ b/pkg/setup/node.go
@@ -67,14 +67,17 @@ func (sn *Node) Serve(ctx context.Context) error {
 			return err
 		}
 		go func(tp transport.Transport) {
-			if err := sn.serveTransport(tp); err != nil {
+			if err := sn.serveTransport(ctx, tp); err != nil {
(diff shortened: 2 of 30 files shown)


evanlinjin:
I'm thinking about the following renaming:
- `Transport` -> `Connection` (ideally we can just use `net.Conn`).
- `ManagedTransport` -> `Transport`.

evanlinjin:
Transport logic is still faulty.

pkg/router/route_manager.go
```diff
@@ -35,6 +35,10 @@ func (rm *routeManager) GetRule(routeID routing.RouteID) (routing.Rule, error) {
 		return nil, errors.New("unknown RouteID")
 	}
 
+	if len(rule) < 13 {
```
nkryuchkov: What do you think of extracting `13` to a constant?

pkg/setup/node.go
```diff
@@ -67,14 +67,17 @@ func (sn *Node) Serve(ctx context.Context) error {
...
+			if err := sn.serveTransport(ctx, tp); err != nil {
 				sn.Logger.Warnf("Failed to serve Transport: %s", err)
 			}
 		}(tp)
 	}
 }
 
-func (sn *Node) serveTransport(tr transport.Transport) error {
+func (sn *Node) serveTransport(ctx context.Context, tr transport.Transport) error {
+	ctx, cancel := context.WithTimeout(ctx, time.Second*30)
```
nkryuchkov: What do you think of extracting a timeout constant here?

pkg/setup/protocol.go
```diff
@@ -130,35 +132,53 @@ func DeleteRule(p *Protocol, routeID routing.RouteID) error {
...
+func LoopClosed(ctx context.Context, p *Protocol, ld routing.LoopData) error {
 	if err := p.WritePacket(PacketLoopClosed, ld); err != nil {
 		return err
 	}
-	return readAndDecodePacket(p, nil)
+	return readAndDecodePacketWithTimeout(ctx, p, nil)
+}
+
+func readAndDecodePacketWithTimeout(ctx context.Context, p *Protocol, v interface{}) error {
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
```
nkryuchkov: I'd extract a constant here too.

pkg/transport/managed_transport.go
```diff
@@ -2,107 +2,351 @@ package transport
...
+			break
+		}
+		select {
+		case <-mt.done:
+			return nil, ErrNotServing
+		case <-mt.connCh:
+		}
+	}
+
+	h := make(routing.Packet, 6)
```
nkryuchkov: Can we extract `6` here?

nkryuchkov (approved):
Looks awesome! I'd extract some constant values to improve the readability, but it's up to you to decide.

PR #510, opened 2019-08-12
Added pty flags to ssh CLI 

ivcosla:
Added cli flag to call RUN with custom PTY options

### DIFF
--- a/cmd/therealssh-cli/commands/root.go
+++ b/cmd/therealssh-cli/commands/root.go
@@ -22,12 +22,13 @@ import (
 )
 
 var (
-	rpcAddr string
-	ptyMode bool
-	ptyRows uint16
-	ptyCols uint16
-	ptyX    uint16
-	ptyY    uint16
+	rpcAddr       string
+	ptyMode       bool
+	ptyRows       uint16
+	ptyCols       uint16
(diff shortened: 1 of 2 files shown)


cmd/therealssh-cli/commands/root.go
```diff
@@ -117,6 +129,63 @@ var rootCmd = &cobra.Command{
...
+		log.Fatal(err)
+	}
+
+	conn, err := net.DialUnix("unix", nil, &net.UnixAddr{Name: socketPath, Net: "unix"})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	b := make([]byte, ptyBufferSize)
+	_, err = conn.Read(b) // nolint
```
nkryuchkov: Is there any linter error here if the comment is removed?

ivcosla: damn, forgot to remove it. Had another function before for reading, which an error that needed to be discarded.

nkryuchkov (approved):

evanlinjin (approved):

aussieboy88 (commented):
What do I do I dont understand any of this but i really want to learn

PR #519, opened 2019-08-18
Make route establishment asynchronous.

### LINKED ISSUE #506: [ROUTING] Make SetupNode create routes asynchronously.
## Context
The Setup Node currently creates routes by synchronously delivering routing rules to visor nodes. If rules can be delivered concurrently, it will greatly increase the speed of route setup.

Darkren:
Fixes #506 

Made route establishment procedure asynchronous. 
- Route establishment procedure was divided into 2 steps:
    1. Request registration ID
    2. Apply rule
- Rule was extended to hold `registration ID` field
- The route establishment procedure itself was put in goroutines (one for each `visor` node applying rule). Any error during route setup cancels the whole procedure
- Tests were adjusted


pkg/router/route_manager.go
```diff
@@ -214,24 +214,56 @@ func (rm *routeManager) RemoveLoopRule(loop routing.Loop) error {
 	return nil
 }
 
-func (rm *routeManager) addRoutingRules(data []byte) ([]routing.RouteID, error) {
+<<<<<<< HEAD
```
evanlinjin: Please fix.

pkg/router/route_manager.go
```diff
@@ -214,24 +214,56 @@ func (rm *routeManager) RemoveLoopRule(loop routing.Loop) error {
...
+func (rm *routeManager) Serve(rwc io.ReadWriteCloser) error {
```
evanlinjin: I've renamed this to `handleSetupConn` and `Serve()` should run a loop. Please check: https://github.com/skycoin/skywire/blob/mainnet-milestone1/pkg/router/route_manager.go

pkg/setup/node.go
```diff
@@ -204,49 +201,107 @@ func (sn *Node) createRoute(ctx context.Context, expireAt time.Time, route routi
...
+
+	// chan to receive the resulting route ID from a goroutine
+	resultingRouteIDCh := make(chan routing.RouteID, 2)
+
+	// context to cancel rule setup in case of errors
+	ctx, cancel := context.WithCancel(context.Background())
 	for idx := len(r) - 1; idx >= 0; idx-- {
-		hop := &Hop{Hop: route[idx]}
-		r[idx] = hop
+		var regIDChIn, regIDChOut chan routing.RouteID
```
evanlinjin: For the above, can you please rename `idx` to `i`. Thanks!

pkg/setup/node.go
```diff
@@ -204,49 +201,107 @@ func (sn *Node) createRoute(ctx context.Context, expireAt time.Time, route routi
...
+	}
+	copy(r[1:], route)
 
 	initiator := route[0].From
+
+	// indicate errors occurred during rules setup
+	rulesSetupErrs := make(chan error, len(r))
+	// regIDsCh is an array of chans used to pass the requested registration IDs around the goroutines.
+	// We do it in a fan fashion here. We create as many goroutines as there are rules to be applied.
+	// Goroutine[idx] requests visor node for a registration ID. It passes this registration ID through a chan to
```
evanlinjin: `idx` -> `i` (as it's represented typically in Golang).

pkg/setup/node.go
```diff
@@ -204,49 +201,107 @@ func (sn *Node) createRoute(ctx context.Context, expireAt time.Time, route routi
...
+	// context to cancel rule setup in case of errors
+	ctx, cancel := context.WithCancel(context.Background())
 	for idx := len(r) - 1; idx >= 0; idx-- {
-		hop := &Hop{Hop: route[idx]}
-		r[idx] = hop
+		var regIDChIn, regIDChOut chan routing.RouteID
+		if idx > 0 {
+			regIDChOut = regIDsCh[idx-1]
+		}
+		var nextTransport uuid.UUID
```
evanlinjin: rename is `nextTpID`

evanlinjin (changes requested):
Good stuff. Excited for these changes. Please check my comments, more is coming.

pkg/routing/rule.go
```diff
@@ -101,6 +101,18 @@ func (r Rule) LocalPort() Port {
 	return Port(binary.BigEndian.Uint16(r[48:]))
 }
 
+// RegistrationID returns route ID which will be used to register this rule within
+// the visor node.
+func (r Rule) RegistrationID() RouteID {
```
evanlinjin: What was the intention behind creating a separate name for `RouteID`? How is this different to the `RouteID` which is used as the key to routing rules?

pkg/setup/node.go
```diff
@@ -204,49 +201,107 @@ func (sn *Node) createRoute(ctx context.Context, expireAt time.Time, route routi
 	}
 
 	sn.Logger.Infof("Creating new Route %s", route)
-	r := make([]*Hop, len(route))
+
+	r := make(routing.Route, len(route)+1)
```
evanlinjin, line 232: Why have we created a second `routing.Route` variable?

pkg/setup/node.go
```diff
@@ -204,49 +201,107 @@ func (sn *Node) createRoute(ctx context.Context, expireAt time.Time, route routi
...
+	// close chan to avoid leaks
+	close(rulesSetupErrs)
+	for _, ch := range regIDsCh {
+		close(ch)
 	}
+	if rulesSetupErr != nil {
+		return 0, rulesSetupErr
+	}
+
+	routeID := <-resultingRouteIDCh
```
evanlinjin, line 325: From my understanding, this may lead to hang. Have a select statement with `ctx.Done()` maybe.

evanlinjin (changes requested):
As a general comment for variable names: Golang style generally prefers shorter variable names.
- Public key: `pk`
- Route: `rt`.
- Route ID: `rtID`
- Transport: `tp`
- Transport ID: `tpID`
- Index: `i`
- initiator: `init`
- responder: `resp`

(We can discuss each of these cases, but this is what I have been following with my own naming conventions).

Also it will be good to have the setup procedure documented. For example, I have no idea why `// adding rule for initiator must result with a route ID` :sweat_smile:

Other than the comments I have made, the code looks good. The next steps is testing and to ensure that it works in interactive environment.

Good stuff!

pkg/router/route_manager.go
evanlinjin: And get rid of the `<<<<<<< HEAD`s :smile:

pkg/setup/node.go
Darkren: @evanlinjin I tried to preserve namings as much as I could, probably not the best idea. I'll change this

pkg/routing/rule.go
Darkren: @evanlinjin There is already route ID there. It servers as a `respRouteID` for app rules and as a `nextRouteID` for forward rules. It's not the ID which is used as a key to store this particular rule inside a visor node. That was the biggest trick there. We make a first request to visor node to request a free route ID for the rule. We use this route ID as registration ID for this rule and we use the same route ID as a `nextRouteID/respRouteID` for the previous visor node along the route. That's why I decided to distinguish these IDs, with the same name it becomes hell to understand what's going on there

pkg/setup/node.go
Darkren: @evanlinjin I'm not sure why it'd been here before, while doing these changes I realised that it's not really needed in a synchronous version. But I needed one to start the goroutines simpler and avoid code duplication. And also not to interfere with the original contents of the route, to avoid the side effect of the func

pkg/setup/node.go
Darkren: Much appreciate, will do

pkg/routing/rule.go
evanlinjin: @Darkren thanks for the explanation, that makes sense. However, I'm still confused as to why we need to modify the `routing.Rule` type?

pkg/setup/node.go
evanlinjin: @Darkren let's make a `"TODO: I-don't-know-why-this-is-here` comment.

evanlinjin: I'm being serious btw.

pkg/routing/rule.go
Darkren: @evanlinjin Well, when we get the free route ID from the visor node, we need to send it the rule and tell which route ID should be used to register this rule. Modifying the rule type was straightforward and the simplest solution with regards to network load. From the visor's perspective we need to deal with a lot of requests simultaneously, if we receive a lot of rule at once, we need to know IDs for these rules

pkg/setup/node.go
Darkren: @evanlinjin I actually know the purpose of this array now, I'm running goroutines below, one goroutine for each element of this array. In mainnet we have a loop over this route and apply rule on each iteration. After this loop we apply a rule for the initiator node. In this implementation there is no necessity to have this a separate step imo, that's why I put it into the loop itself into the dedicated goroutine. The code for initiator is just the same as for every other node, so, making this a separate step is just duplicating the code

pkg/routing/rule.go
evanlinjin: So when a visor node receives a rule, it checks the Registration ID to know which request is satisfied? Would it be better to call it `RequestRouteID`?

Darkren: @evanlinjin Yes. I may call it a `RequestRouteID`, np, it's a subject for discussion. My main intent was to disinguish it from the original `routeID` inside a rule, the naming can be whatever fits best

evanlinjin: Fair enough, I see your point. Let's rename to `RequestRouteID` as the idea of "Registration" may be confused with other things.

pkg/setup/node.go
evanlinjin: @Darkren do you mean we can get rid of it?

Darkren: @evanlinjin Well, yes, but we'll still be looping over the original route plus one iteration for the initiating node

evanlinjin: Would it be safe to modify the original `route` variable? Remember to put a comment above such a modification step (if needed) to explain the purpose of modification.

Darkren: @evanlinjin I'll need to check the code base to be certain about safety. Also even if it's safe now, we can't guarantee the same for future changes. That's why I created new variable for this purpose. I think I can get rid of it and run a separate goroutine manually without modifying the original route, I'm not a fan of this, consequences may be tough. I'll see what I can do

evanlinjin: @Darkren I see what you mean. Let's do this:
```golang
route = append(&routing.Hop{
		Transport: route[0].Transport,
		To:        route[0].From,
	}, route[1:]...)
```

evanlinjin: I see, sorry I deleted my previous comment as it was a mistake. Let's create a comment above it detailing the intentions behind it.

Darkren: @evanlinjin I left some comments there and also heavily commented the whole function. Please check it out and tell me if it needs any further clarifications

pkg/setup/node.go
Darkren: @evanlinjin I create inner context from the ctx passed to the func. In case the outer ctx gets cancelled, the inner one gets too. In this case the whole procedure gets cancelled. Right after the goroutines stop there is a check for any errors along the way. If there are any - func returns immideately. If there was none, we execute this line you mentioned above. At this part of execution the value is guaranteed to be there. I added more comments at this part, please check if they're sufficient

evanlinjin: @Darkren Thanks for the update, will look after meeting.

pkg/setup/node.go
```diff
@@ -198,55 +195,147 @@ func (sn *Node) createLoop(ctx context.Context, ld routing.LoopDescriptor) error
...
-		var rule routing.Rule
-		if idx == len(r)-1 {
-			rule = routing.AppRule(expireAt, 0, initiator, lport, rport)
+
+	// add the initiating node to the start of the route. We need to loop over all the visor nodes
+	// along the route to apply rules including the initiating one
+	r := make(routing.Route, len(route)+1)
+	r[0] = &routing.Hop{
+		Transport: route[0].Transport,
+		To:        route[0].From,
```
evanlinjin, line 235: Why do we need to swap the `To` and `From` fields and get rid of the `To` field for the initiating element?

pkg/setup/node.go
```diff
@@ -198,55 +195,147 @@ func (sn *Node) createLoop(ctx context.Context, ld routing.LoopDescriptor) error
...
+	reqIDsCh := make([]chan routing.RouteID, 0, len(r))
+	for range r {
+		reqIDsCh = append(reqIDsCh, make(chan routing.RouteID, 2))
+	}
+
+	// chan to receive the resulting route ID from a goroutine
+	resultingRouteIDCh := make(chan routing.RouteID, 2)
+
+	// context to cancel rule setup in case of errors
+	cancellableCtx, cancel := context.WithCancel(ctx)
```
evanlinjin, line 258:
```golang
ctx, cancel := context.WithCancel(ctx)
```
Will be fine.

pkg/setup/node.go
Darkren: @evanlinjin That's because we're iterating over `visor` nodes we send requests to. For each node we use the `To` field. But the initiating node stands aside, there is no initiating node in the `route`. So, we expand the original rotue by adding the initiating node first in the list and we fill `To` with `From` from the first `visor` node along the route. This corresponds to the initiating node

pkg/setup/node.go
```diff
@@ -198,55 +195,147 @@ func (sn *Node) createLoop(ctx context.Context, ld routing.LoopDescriptor) error
...
+			rulesSetupErrs <- nil
+		}(i, r[i].To, rule, reqIDChIn, reqIDChOut)
 	}
 
-	rule := routing.ForwardRule(expireAt, r[0].routeID, r[0].Transport)
-	routeID, err := sn.setupRule(ctx, initiator, rule)
-	if err != nil {
-		return 0, fmt.Errorf("rule setup: %s", err)
+	var rulesSetupErr error
+	var cancelOnce sync.Once
```
evanlinjin, line 302:
```golang
func main() {
	ctx := context.Background()
	ctx, cancel := context.WithCancel(ctx)

	const n = 1000000
	wg := new(sync.WaitGroup)
	wg.Add(n)

	for i := 0; i < n; i++ {
		go func(i int) {
			cancel()
			wg.Done()
		}(i)
	}

	<-ctx.Done()
	fmt.Println("Done!")
	wg.Wait()
}
```
I just tested `ctx`, and it seems that this is not needed.

pkg/setup/node.go
Darkren: @evanlinjin this way we're overwriting the new context to the original passed to the whole func

Darkren: @evanlinjin I'll test it myself once again, but I feel like the `cancellationCtx` is used only within the func to stop the internal process. And if we call `cancel`, the outer processes using the original `ctx` won't be affected. But if the original `ctx` was also created `WithCancel` and some outer code calls the corresponding `cancel` func for the outer `ctx` the whole func will be cancelled correctly. Well, at least I counted on that behavior, I'll check it out

Issue #226, opened 2019-03-18
[mainnet-manager] Manager Node backend additions.


evanlinjin:
## Description

This is a parent issue of multiple sub-issues/sub-PRs. This is provided here for a clear overview of tasks.

**From @jdknives on Discord:**

> Below the functionality needed for the manager backend (given how the frontend works, and most of those functions make sense imo): :point_down:
> Backend:
> - health endpoint (or directly in nodes resource) to report status of the node
> - uptime
> - ip of the node (for labeling and easier debugging instead of reading long keys)
> - version (of the node software)
> - loops endpoint
> - login (password protection to serve the manager on the web) + logout endpoint
> - change password endpoint (dont let the user use other endpoints if he hasnt changed the default pw first)
> - log/output from apps
> - native apps dont have endpoints yet (for controlling ssh...), once they do, there could also be a history for last used startup parameters (like whitelist for ssh server, nodes for proxy)
> - old manager had a "connect to node" (proxy app) with list of all discovered nodes and their location (country)
> - terminal (security: https://github.com/skycoin/skywire/issues/71)
> - check for updates and update
> - reboot node
> - some help texts? ex. what is exit node toggle for...

## Tasks
- [x] Move RPC implementation back to `/pkg/node` module.
- [x] Implement login/logout endpoints (authboss?).
- [ ] Node uptime endpoint.
- [ ] Health endpoint.
- [ ] Uptime tracking service, manager and app node communication.
- [x] IP endpoints.
- [x] Versioning endpoints.
- [x] Loops endpoints.
- [ ] Manager access to logging from apps and node.
- [ ] Manager access to updater + reboot.
- [x] Manager endpoint: Call discovery server with "connect to node...".
- [ ] Manager access to terminal.

jdknives:
@evanlinjin note that uptime only requires the time elapsed since the last connection to the discovery (ap4y suggested implementing uptime tracking in transport discvory). The total uptime is calculated by the monitor, although we could improve on that. 

Could you elaborate on why you want to move the RPC implementation back to /pkg/node?

evanlinjin:
> @evanlinjin note that uptime only requires the time elapsed since the last connection to the discovery (ap4y suggested implementing uptime tracking in transport discvory). The total uptime is calculated by the monitor, although we could improve on that.
> 
> Could you elaborate on why you want to move the RPC implementation back to /pkg/node?

[...]

PR #521, opened 2019-08-20
Manager node backend additions.

ivcosla:
Implements #226


cmd/apps/skychat/chat.go
```diff
@@ -30,12 +31,15 @@ var (
...
 	connsMu   sync.Mutex
+	log       *logging.MasterLogger
 )
 
 func main() {
 	flag.Parse()
-
-	a, err := app.Setup(&app.Config{AppName: "skychat", AppVersion: "1.0", ProtocolVersion: "0.0.1"})
+	appName := "skychat"
+	fmt.Println(os.Args)
```
nkryuchkov: Does this line have any purpose? Looks like it may be a forgotten debug print.

cmd/apps/skychat/chat.go
```diff
@@ -9,14 +9,15 @@ import (
...
-	"log"
 	"net"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 
 	"github.com/skycoin/dmsg/cipher"
 
+	"github.com/skycoin/skycoin/src/util/logging"
```
nkryuchkov, line 18: Could you please sort imports here?

cmd/apps/skychat/chat.go
```diff
@@ -30,12 +31,15 @@ var (
...
+	log       *logging.MasterLogger
 )
 
 func main() {
 	flag.Parse()
-
-	a, err := app.Setup(&app.Config{AppName: "skychat", AppVersion: "1.0", ProtocolVersion: "0.0.1"})
+	appName := "skychat"
+	fmt.Println(os.Args)
+	log, _ = app.NewLogger(appName, os.Args)
```
nkryuchkov: Could we handle a probable error here?

cmd/skywire-cli/commands/node/app.go
```diff
@@ -82,3 +83,21 @@ var setAppAutostartCmd = &cobra.Command{
 		fmt.Println("OK")
 	},
 }
+
+var appLogsSince = &cobra.Command{
+	Use:   "app-logs-since <name> <timestamp>",
+	Short: "Gets logs from given app since RFC3339Nano-formated timestamp",
```
nkryuchkov: Could you please change `formated` to `formatted`?

pkg/app/log.go
```diff
@@ -0,0 +1,51 @@
+package app
+
+import (
+	"github.com/skycoin/skycoin/src/util/logging"
```
nkryuchkov: Could you please format imports in this file?

pkg/app/log_store.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+	if err != nil && !strings.Contains(err.Error(), bbolt.ErrBucketExists.Error()) {
+		return nil, err
+	}
+
+	return &boltDBappLogs{db, b}, nil
+}
+
+// Write implements io.Writer
+func (l *boltDBappLogs) Write(p []byte) (int, error) {
```
nkryuchkov, line 68: What do you think of checking the length of `p` here and return an error if the function is misused?

pkg/app/log_store.go
```diff
@@ -0,0 +1,137 @@
...
+			return iterateFromBeginning(c, parsedTime, &logs)
+		}
+		c.Seek(parsedTime)
+		return iterateFromKey(c, &logs)
+	})
+
+	return logs, err
+}
+
+func iterateFromKey(c *bbolt.Cursor, logs *[]string) error {
```
nkryuchkov: Could we rewrite `iterateFromKey` and `iterateFromBeginning` in a way that it returns logs along with an error instead of appending them by pointer?

pkg/app/log_store.go
```diff
@@ -0,0 +1,137 @@
...
+	for k, v := c.Next(); k != nil; k, v = c.Next() {
+		t, err := bytesToTime(k)
+		if err != nil {
+
```
nkryuchkov: Could you please remove the empty line here?

pkg/app/log_store_test.go
```diff
@@ -0,0 +1,56 @@
+package app
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/stretchr/testify/require"
```
nkryuchkov, line 10: Could you please sort imports here?

pkg/app/log_test.go
```diff
@@ -0,0 +1,71 @@
+package app
+
+import (
+	"fmt"
+	"github.com/skycoin/skycoin/src/util/logging"
```
nkryuchkov: Could you please sort imports here?

pkg/app/log_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestWriteLog(t *testing.T) {
+	r, w := io.Pipe()
+
+	l := logging.NewMasterLogger()
+	l.SetOutput(w)
+	l.Logger.Formatter.(*logging.TextFormatter).TimestampFormat = time.RFC3339Nano
+	c := make(chan []byte)
+
+	go func() {
+		b := make([]byte, 51)
```
nkryuchkov: Why does `b` have the size of `51` here?

pkg/hypervisor/hypervisor.go
```diff
@@ -150,6 +152,49 @@ func (m *Node) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+// VisorHealth represents a node's health report attached to it's pk for identification
+type VisorHealth struct {
+	PK     cipher.PubKey `json:"pk"`
+	Status int           `json:"status"`
+	*visor.HealthInfo
+}
+
+// provides summary of health information for every visor
+func (m *Node) getHealth() http.HandlerFunc {
+	healthStatuses := make([]*VisorHealth, len(m.nodes))
```
nkryuchkov: You probably meant `healthStatuses := make([]*VisorHealth, 0, len(m.nodes))` here.

pkg/visor/rpc.go
```diff
@@ -3,11 +3,15 @@ package visor
 import (
 	"context"
 	"errors"
+	"path/filepath"
 	"time"
 
+	"net/http"
```
nkryuchkov: Could you please sort imports here?

pkg/visor/rpc_client.go
```diff
@@ -271,6 +306,22 @@ func (mc *mockRPCClient) Summary() (*Summary, error) {
 	return &out, err
 }
 
+// Health implements RPCClient
+func (mc *mockRPCClient) Health() (*HealthInfo, error) {
+	hi := &HealthInfo{
+		TransportDiscovery: 200,
```
nkryuchkov: `http.StatusOK` might be more readable here

pkg/app/log.go
```diff
@@ -0,0 +1,51 @@
...
+	"io"
+	"time"
+)
+
+// NewLogger is like (a *App) LoggerFromArguments but with appName as parameter, instead of
+// getting it from app config
+func NewLogger(appName string, args []string) (*logging.MasterLogger, []string) {
```
evanlinjin: Why would we use this as apposed to `(*App).LoggerFromArguments()`?

pkg/app/log_store.go
evanlinjin: @nkryuchkov How do you suggest we do this? And how should it be used in `LogsSince`?

pkg/app/log_test.go
```diff
@@ -0,0 +1,71 @@
...
+		c <- b
+	}()
+	l.Println("foo")
+
+	res := <-c
+	ti := res[1:36]
+
+	pt, err := time.Parse(time.RFC3339Nano, string(ti))
+	if err != nil {
+		t.Fail()
```
evanlinjin: We need to use `t.FailNow()` or else it does not return. A better method would be to use `require.NoError()` to be consistent.

pkg/app/log_test.go
evanlinjin: From my understanding, this is just an arbitrary value to contain the reads. If not, @ivcosla, please create a constant.

pkg/app/log_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/skycoin/skycoin/src/util/logging"
+	"github.com/stretchr/testify/require"
+	"io"
+	"io/ioutil"
+	"os"
+	"testing"
+	"time"
+)
+
+func TestWriteLog(t *testing.T) {
```
evanlinjin: Some comments to describe what is being tested will be helpful.

pkg/app/log_test.go
```diff
@@ -0,0 +1,71 @@
...
+	res := <-c
+	ti := res[1:36]
+
+	pt, err := time.Parse(time.RFC3339Nano, string(ti))
+	if err != nil {
+		t.Fail()
+	}
+
+	fmt.Println("t in unix nano", pt.UnixNano())
+	fmt.Printf("%#v", string(res))
```
evanlinjin: Should we check this output?

pkg/hypervisor/hypervisor.go
evanlinjin: We should probably write tests for these endpoints.

evanlinjin (changes requested):
Good stuff. Some comments.

cmd/apps/skychat/chat.go
ivcosla: Yes just a print, that I should remove when work on the PR is finished.

cmd/apps/skychat/chat.go
ivcosla: It's not an error. Second argument is `[]string` which is `os.Args` with the first argument stripped to make it more transparent to use for the user. But this function signature requires rethinking, I'm not happy with it. If there is an error inside it panics.

Ideally, I think we should return 3 arguments during `app.Setup`-> `(*App, *logging.MasterLogger, error)`. The idea is that the user must always get the logger to prevent scenarios in which he would forgot to create a logger and thus more likely to use it. I don't want to embed it in `*App` since the logger should be available at app setup time, in case that an error is returned an app is nil the logger should still be used to print the fatal error so the hypervisor can fetch it.

pkg/app/log_store.go
ivcosla: I'm not sure about this, since `Write` is called from `*logging.MasterLogger`I'm not sure which misused there might be. An empty log maybe?, is it to be discarded or still persisted for hypervisor to fetch?.

But I agree should be checked in case that there might be a misuse. Do you have any other scenario in mind?

ivcosla:
There are linter errors preventing the PR to pass tests on travis. Those errors belong to other packages and should be fixed in a different PR.

nkryuchkov:
Linter errors are fixed in https://github.com/skycoin/skywire/pull/537

cmd/apps/skychat/chat.go
nkryuchkov: @ivcosla Sorry, I didn't pay enough attention here and assumed that it returns a value and an error.

pkg/app/log_store.go
nkryuchkov: @ivcosla It may be misused if `len(p) < 37`. I think this may happen if logger settings omit timestamp or specify another log format. Not sure if it may happen, but this may improve robustness. I'd add the following:
```suggestion
func (l *boltDBappLogs) Write(p []byte) (int, error) {
	if len(p) < 37 {
		return 0, io.ErrShortBuffer
	}
```

pkg/app/log_store.go
nkryuchkov: @evanlinjin I would write the function this way:
```go
func iterateFromKey(c *bbolt.Cursor) ([]string, error) {
	logs := make([]string, 0)
	for k, v := c.Next(); k != nil; k, v = c.Next() {
		logs = append(logs, string(v))
	}
	return logs, nil
}
```
And then append the returned value to the `logs` variable outside the function.

On the one hand, it would decrease the performance a bit. On the other hand, I think it would be more readable.

evanlinjin: I think that is better.

pkg/app/log_store.go
```diff
@@ -0,0 +1,163 @@
...
+// Write implements io.Writer
+func (l *boltDBappLogs) Write(p []byte) (int, error) {
+	db, err := bbolt.Open(l.dbpath, 0600, nil)
+	if err != nil {
+		return 0, err
+	}
+	defer func() {
+		err := db.Close()
+		if err != nil {
+			panic(err)
```
nkryuchkov, line 81: Is there a reason why panicking instead of returning an error is preferred here and in `Store` method?

cmd/skywire-cli/commands/node/app.go
```diff
@@ -85,6 +87,31 @@ var setAppAutostartCmd = &cobra.Command{
...
 }
 
+var appLogsSinceCmd = &cobra.Command{
+	Use:   "app-logs-since <name> <timestamp>",
+	Short: "Gets logs from given app since RFC3339Nano-formated timestamp. \"beginning\" is a special timestamp to fetch all the logs",
+	Args:  cobra.MinimumNArgs(2),
+	Run: func(_ *cobra.Command, args []string) {
+		var t time.Time
+
+		if args[1] == "beginning" {
```
nkryuchkov, line 97: What do you think of extracting the `"beginning"` constant here?

pkg/hypervisor/hypervisor.go
```diff
@@ -151,6 +156,55 @@ func (m *Node) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		select {
+		case res := <-resCh:
+			if res.err != nil {
+				vh.Status = http.StatusInternalServerError
+			} else {
+				vh.HealthInfo = res.h
+				vh.Status = http.StatusOK
+			}
+			httputil.WriteJSON(w, r, http.StatusOK, vh)
+			return
```
nkryuchkov: Could we get rid of it? Looks redundant.

pkg/hypervisor/hypervisor.go
```diff
@@ -151,6 +156,55 @@ func (m *Node) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		}
+	})
+}
+
+// getUptime gets given node's uptime
+func (m *Node) getUptime() http.HandlerFunc {
+	return m.withCtx(m.nodeCtx, func(w http.ResponseWriter, r *http.Request, ctx *httpCtx) {
+		u, err := ctx.RPC.Uptime()
+		if err != nil {
+			httputil.WriteJSON(w, r, http.StatusInternalServerError, err)
```
nkryuchkov, line 201: I think we need a return statement here, because `httputil.WriteJSON(w, r, http.StatusOK, u)` would be also executed.

evanlinjin:
What's the situation with this PR?

ivcosla:
It can be merged after conflicts are fixed

PR #522, opened 2019-08-21
Implement arbitrary command execution on visor from hypervisor

nkryuchkov:
Part of #226 

Execution of arbitrary commands on visor from hypervisor is needed to run update scripts on the visor.

Done:
- RPC implementation
- RPC usage in CLI
- RPC usage in API


evanlinjin:
Technically we can also interact with Apps running on visor node in this manner. So we are pretty much killing two birds with one stone.

evanlinjin (approved):

PR #512, opened 2019-08-14
[TODO:NeedRebase] Finish renaming to dmsg

nkryuchkov:
Closes #502 

Done:
- Rename `messaging` to `dmsg`

PR #526, opened 2019-08-22
[M2] Finish renaming to dmsg

### LINKED ISSUE #502: Finish renaming to dmsg
While reviewing the codebase, I found multiple instances of old naming conventions. This affects for instance

- our config files (it still contains references to messaging and node)

- /cmd (contains references to messaging)

- the specs

nkryuchkov:
Closes #502 

Copy of https://github.com/skycoin/skywire/pull/512 for `mainnet-milestone1` branch.

Done:
- Rename `messaging` to `dmsg`


evanlinjin (commented):
We need an accompanying PR into `skywire-services` that ensures that interactive environment still compiles and runs as expected, as the renaming of binaries will cause issues.

evanlinjin:
I have a change of mind. We should probably do this after all the other `mainnet-milestone1` PRs are merged.

PR #527, opened 2019-08-22
Change rule expiry to keep-alive

### LINKED ISSUE #513: [BUG] Route expiration is buggy.
Sometimes routing rules expire even when the routing rule is constantly used.

## To reproduce
Run the messaging interactive test environment for a while.

Darkren:
Fixes #513 

Modified rule to contain `KeepAlive` field instead of `ExpiresAt` also adjusting the client code and tests


pkg/router/managed_routing_table.go
```diff
@@ -23,21 +27,41 @@ func manageRoutingTable(rt routing.Table) *managedRoutingTable {
...
 }
 
-func (rt *managedRoutingTable) Rule(routeID routing.RouteID) (routing.Rule, error) {
+func (rt *managedRoutingTable) AddRule(rule routing.Rule) (routing.RouteID, error) {
+	routeID, err := rt.Table.AddRule(rule)
+	if err != nil {
+		return 0, err
+	}
+
 	rt.mu.Lock()
```
evanlinjin: We should also lock the `rt.Table.AddRule()` to ensure the entire operation is atomic (as there is a very small possibility that the added rule is deleted immediately by `Cleanup`).

pkg/router/router.go
```diff
@@ -23,8 +23,8 @@ import (
 )
 
 const (
-	// RouteTTL is the default expiration interval for routes
-	RouteTTL = 2 * time.Hour
+	// RouteKeepAlive is the default expiration interval for routes
+	RouteKeepAlive = 2 * time.Hour
```
evanlinjin: Rename to `DefaultRouteKeepAlive`.

pkg/visor/rpc_client.go
```diff
@@ -199,7 +199,7 @@ func NewMockRPCClient(r *rand.Rand, maxTps int, maxRules int) (cipher.PubKey, RP
 		log.Infof("tp[%2d]: %v", i, tps[i])
 	}
 	rt := routing.InMemoryRoutingTable()
-	ruleExp := time.Now().Add(time.Hour * 24)
+	ruleKeepAlive := 24 * time.Hour
```
evanlinjin: Use `router.DefaultRouteKeepAlive`.

evanlinjin (changes requested):
Good work. Just some comments here and there.

pkg/router/managed_routing_table.go
Darkren: @evanlinjin fair. Fixed this one by putting `AddRule` and `DeleteRules` calls under the lock, please check out if it's appropriate

evanlinjin: We need to move `rt.mu.Lock()` to line `31`. There is still a possibility that `Cleanup` will trigger on execution hits line `32`, hence, deleting the rule as soon as it's created.

Darkren: @evanlinjin I'm sorry, made a stupid move. I made `Rule` atomic instead of `AddRule` just confused methods. However, I made `AddRule` atomic now. And the question is, should `Rule` also be atomic as it is now?

evanlinjin: Ahh I see, that makes sense. Yeah sure, let's keep it as is. Now we have consistency across the methods.

pkg/router/managed_routing_table.go
```diff
@@ -23,29 +27,58 @@ func manageRoutingTable(rt routing.Table) *managedRoutingTable {
...
+	defer rt.mu.Unlock()
+
 	err := rt.RangeRules(func(routeID routing.RouteID, rule routing.Rule) bool {
-		if rule.Expiry().Before(time.Now()) {
-			if lastActivity, ok := rt.activity[routeID]; !ok || time.Since(lastActivity) > routeKeepalive {
-				expiredIDs = append(expiredIDs, routeID)
-			}
+		if rt.ruleIsTimedOut(routeID, rule) {
+			expiredIDs = append(expiredIDs, routeID)
```
evanlinjin, line 70: We need to have `delete(rt.activity, routeID)` somewhere.

evanlinjin (changes requested):
Looks good. Just one thing.

pkg/router/managed_routing_table.go
Darkren: Fair

evanlinjin (approved):

PR #533, opened 2019-08-26
Read arbitrary hypervisor config from flag instead of args

### LINKED ISSUE #532: Hypervisor cannot read an arbitrary config passed through args
Hypervisor has two `Cobra` commands: 
- root 
- `gen-config`

If using root and specifying a config file to be used, `Cobra` thinks that the config path is a command name, so arbitrary config does not work now. We need to fix this in a way that it reads config from an arbitrary path using a flag.

nkryuchkov:
Closes #532

### DIFF
--- a/pkg/util/pathutil/configpath.go
+++ b/pkg/util/pathutil/configpath.go
@@ -97,8 +97,9 @@ func HypervisorDefaults() ConfigPaths {
 // - From CLI argument.
 // - From ENV.
 // - From a list of default paths.
+// If argsIndex < 0, args search does not take place.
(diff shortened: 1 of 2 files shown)


pkg/util/pathutil/configpath.go
```diff
@@ -97,8 +97,9 @@ func HypervisorDefaults() ConfigPaths {
 // - From CLI argument.
 // - From ENV.
 // - From a list of default paths.
+// If argsIndex < 0, args search does not take place.
```
evanlinjin: `args` -> `CLI args` to avoid ambiguity.

evanlinjin (approved):
Good work. Just the single comment.

PR #534, opened 2019-08-27
Fix/tcp-transport with correct laddr

ayuryshev:
It's a part of bigger and unfinished tcp-transport PR.

Here:
- Fix of tcp-transport itself - it did not send laddr during Dial so
on Accept on another edge was broken
- Fix of memoryPubKeyTable: reverse search of pk was broken
- FilePubKeyTable - simplified, now composed from memoryPubKeyTable
- tests of tcp-transport - under no_ci tags because they need prepared IP addresses
- .travis-ci now prepares IP aliases for tests

Also: 
Fixed message `can't load package: package github.com/skycoin/skywire: unknown import path "github.com/skycoin/skywire": cannot find module providing package github.com/skycoin/skywire` during commands like `go clean`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,10 @@ install:
   - go get -u github.com/FiloSottile/vendorcheck
   - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $GOPATH/bin v1.17.1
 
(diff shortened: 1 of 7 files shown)


evanlinjin:
@ayuryshev I will close this PR as it seems these changes are already added to #526

PR #536, opened 2019-08-27
Milestone-1 TCP-Transport. 

ayuryshev:
Excluding the last task defined  in https://github.com/skycoin/skywire/issues/501 - everything is completed.
It was already done by most part a long time before.
The most hard part is in the last task.


pkg/snet/network.go
```diff
@@ -39,16 +40,22 @@ type Config struct {
 
 	DmsgDiscAddr string
 	DmsgMinSrvs  int
+
+	LocalTCPAddress string
```
evanlinjin: Rename to `TCPLocalAddr`

pkg/snet/network.go
```diff
@@ -39,16 +40,22 @@ type Config struct {
...
+	PubKeyFile      string
```
evanlinjin: Rename to `TCPPubKeyFile`

pkg/snet/network.go
```diff
@@ -26,6 +26,7 @@ const (
 // Network types.
 const (
 	DmsgType = "dmsg"
+	TCPType  = "tcptr"
```
evanlinjin: Should be `tcp`

pkg/snet/network.go
```diff
@@ -118,9 +133,33 @@ func (n *Network) Listen(network string, port uint16) (*Listener, error) {
...
+				network:  TCPType,
+			}, nil
+		}
 	default:
 		return nil, ErrUnknownNetwork
 	}
+	return nil, nil
+}
+
+func (n *Network) TCP() *TCPFactory {
```
evanlinjin: Put this on line `107` to keep things orderly.

pkg/snet/tcp_transport.go
```diff
@@ -1,9 +1,12 @@
-package transport
+// +build !no_ci
```
evanlinjin: Rename file to `tcp.go`. `Transport` is no longer an interface, but something that wraps `net.Conn`.

pkg/snet/tcp_transport.go
```diff
@@ -101,35 +119,52 @@ func (tr *TCPTransport) Type() string {
...
+func (t *memPKTable) RemoteAddr(remotePK cipher.PubKey) string {
 	return t.entries[remotePK]
 }
 
-func (t *inMemoryPKTable) RemotePK(remoteIP net.IP) cipher.PubKey {
-	for pk, addr := range t.entries {
-		if addr.IP.String() == remoteIP.String() {
-			return pk
-		}
+func (t *memPKTable) RemotePK(address string) cipher.PubKey {
```
evanlinjin: Have this return `(cipher.PubKey, bool)`

pkg/snet/tcp_transport.go
```diff
@@ -74,7 +92,7 @@ func (f *TCPFactory) Local() cipher.PubKey {
 
 // Type returns the Transport type.
 func (f *TCPFactory) Type() string {
-	return "tcp"
+	return "tcp-transport"
```
evanlinjin: `"tcp"` is fine. Better to use the defined constant though.

pkg/snet/tcp_transport.go
evanlinjin: I think it is best to put this in a separate package `/pkg/snet/stcp`.

pkg/snet/tcp_transport.go
```diff
@@ -101,35 +119,52 @@ func (tr *TCPTransport) Type() string {
 
 // PubKeyTable provides translation between remote PubKey and TCPAddr.
 type PubKeyTable interface {
-	RemoteAddr(remotePK cipher.PubKey) *net.TCPAddr
-	RemotePK(remoteIP net.IP) cipher.PubKey
+	RemoteAddr(remotePK cipher.PubKey) string
```
evanlinjin: Why have we decided to use `string` instead of `net.TCPAddr`?

pkg/snet/tcp_transport.go
```diff
@@ -101,35 +119,52 @@ func (tr *TCPTransport) Type() string {
...
+}
+
+type memPKTable struct {
+	entries map[cipher.PubKey]string
+	reverse map[string]cipher.PubKey
 }
 
-type inMemoryPKTable struct {
-	entries map[cipher.PubKey]*net.TCPAddr
+func memoryPubKeyTable(entries map[cipher.PubKey]string) *memPKTable {
```
evanlinjin: Rename to `makeMemPKTable` to be consistent.

pkg/snet/tcp_transport.go
```diff
@@ -101,35 +119,52 @@ func (tr *TCPTransport) Type() string {
...
+	reverse := make(map[string]cipher.PubKey)
+	for k, v := range entries {
+		addr, err := net.ResolveTCPAddr("tcp", v)
+		if err != nil {
+			panic("error in resolving address")
+		}
+		reverse[addr.IP.String()] = k
```
evanlinjin: Why are we using purely IP? Can we just use tcp address with `host:port`?

pkg/snet/tcp_transport.go
```diff
@@ -101,35 +119,52 @@ func (tr *TCPTransport) Type() string {
...
+	}
+	return &memPKTable{entries, reverse}
 }
 
-// InMemoryPubKeyTable returns in memory implementation of the PubKeyTable.
-func InMemoryPubKeyTable(entries map[cipher.PubKey]*net.TCPAddr) PubKeyTable {
-	return &inMemoryPKTable{entries}
+// MemoryPubKeyTable returns in memory implementation of the PubKeyTable.
+func MemoryPubKeyTable(entries map[cipher.PubKey]string) PubKeyTable {
```
evanlinjin: `MakeMemoryPubKeyTable`

evanlinjin (changes requested):

pkg/snet/tcp_transport.go
ayuryshev: It was my first thought too.
But then I remembered/encountered that for dialling side an ephemereal port is assigned. 
In the case when N nodes are dialling to single node that accepts connections - all diallers will have different ephemereal port.
So PK<==>IP  is simplest (and maybe the only practical) decision.

pkg/snet/tcp_transport.go
ayuryshev: Renamed.
About separate package: `net` from stdlib uses only one nesting level and covers all types of TCP/IP. I propose to stick with this example.

pkg/snet/tcp_transport.go
ayuryshev: No reason :)
I just forgot to change signature after merge from mainnet
Although reasons for using address in string forms exists. 
Take a look how  became:
```golang
func (t *memPKTable) RemoteAddr(remotePK cipher.PubKey) *net.TCPAddr {
	rAddr, _ := net.ResolveTCPAddr("tcp", t.entries[remotePK]) // nolint: errcheck
	return rAddr
}
```

pkg/snet/tcp_transport.go
ayuryshev: OK

ayuryshev:
Fixes requested by Evan - done.
Linter errors and failed tests that were before this PR remains the same.

ayuryshev:
Remerged with mainnet-milestone-1. Fixed tests for tcp-transport

PR #537, opened 2019-08-27
Fix linter errors

nkryuchkov:
Right now, TravisCI builds fail for all pull requests on `mainnet-milestone` branch because it contains linter errors. This PR fixes the errors that break builds.

### DIFF
--- a/internal/utclient/client.go
+++ b/internal/utclient/client.go
@@ -12,9 +12,12 @@ import (
 	"net/http"
 
 	"github.com/skycoin/dmsg/cipher"
+	"github.com/skycoin/skycoin/src/util/logging"
 
 	"github.com/skycoin/skywire/internal/httpauth"
 )
+
+var log = logging.MustGetLogger("utclient")
 
 // Error is the object returned to the client when there's an error.
 type Error struct {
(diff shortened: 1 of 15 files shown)


evanlinjin (commented):
@nkryuchkov I'm curious on the reasoning behind logging error instead of ignoring them with a `nolint` comment? For closing connections, sometimes the connection is already closed, and we just need to ensure that they are closed so the error does not need to be handled.

nkryuchkov:
@evanlinjin I wanted us to have more control of what's going on in tests, however, I had no thought that connections might be already closed. What do you think of checking if `err` contains either `nil` or an `already closed` error? I don't like linter suppression and I think it might be more clear if we could handle errors this way.

evanlinjin:
> @evanlinjin I wanted us to have more control of what's going on in tests, however, I had no thought that connections might be already closed. What do you think of checking if `err` contains either `nil` or an `already closed` error? I don't like linter suppression and I think it might be more clear if we could handle errors this way.

Okay good point and good solution. Let's do it that way then. Please make relevant changes.

Darkren (approved):
Feels like a good job to me

PR #543, opened 2019-08-30
[M2] app2 implementation

### LINKED ISSUE #528: [M2] Begin implementation of the 'app2' module.
## Context
The `app2` module aims to improve on and replace the `app` module (eventually). For now, we should use `app2` to implement direct `dmsg` connections to the skywire hypervisor.

**Specifications:** https://github.com/watercompany/skywire-specs/tree/master/proposals/evanlinjin-app2

Darkren:
Fixes #528 

- Added basic structures
- Implemented methods for `HSFrame`


pkg/app2/hsframe.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"encoding/binary"
+	"encoding/json"
+	"errors"
+	"math"
+)
+
+const (
+	HSFrameHeaderLength  = 5
```
evanlinjin: Use `Len` instead of `Length`.

pkg/app2/hsframe.go
```diff
@@ -0,0 +1,110 @@
...
+	HSFrameProcIDLength  = 2
+	HSFrameTypeLength    = 1
+	HSFrameBodyLenLength = 2
+	HSFrameMaxBodyLength = math.MaxUint16
+)
+
+var (
+	// ErrHSFrameBodyTooLong is being returned when the body is too long to be
+	// fit in the HSFrame
+	ErrHSFrameBodyTooLong = errors.New("frame body is too long")
```
evanlinjin: `TooLarge`

evanlinjin (commented):
:page_facing_up:

pkg/app2/client.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+// Client is used by skywire apps.
+type Client struct {
+	PK          cipher.PubKey
+	pid         ProcID
+	sockAddr    string
+	conn        net.Conn
+	session     *yamux.Session
+	logger      *logging.Logger
```
evanlinjin: Let's call this `log` to keep things concise.

pkg/app2/client.go
```diff
@@ -0,0 +1,178 @@
...
+	lm          *listenersManager
+	isListening int32
+}
+
+// NewClient creates a new Client. The Client needs to be provided with:
+// - localPK: The local public key of the parent skywire visor.
+// - pid: The procID assigned for the process that Client is being used by.
+// - sockAddr: The socket address to connect to Server.
+func NewClient(localPK cipher.PubKey, pid ProcID, sockAddr string, l *logging.Logger) (*Client, error) {
```
evanlinjin: Make the `*logging.Logger` as first input. We will try stick to this convention for the rest of our structures.

pkg/app2/client.go
```diff
@@ -0,0 +1,178 @@
...
+		PK:       localPK,
+		pid:      pid,
+		sockAddr: sockAddr,
+		conn:     conn,
+		session:  session,
+		lm:       lm,
+	}, nil
+}
+
+func (c *Client) Dial(addr routing.Addr) (net.Conn, error) {
```
evanlinjin: For each of these, I would put the `handshake` step in it's own set of functions.

```golang
func dmsgDialHS(conn net.Conn, pid ProcID, /*other variables needed for handshake*/) error
```

Instead of outputs, we can have pointer inputs.

This way:
1. It improves the testability of the code.
2. It groups each communication exchange together; `dmsgDialHS(), handleDmsgDialHS(), dmsgListenHS(), handleDmsgListenHS()` etc.

pkg/app2/client.go
```diff
@@ -0,0 +1,178 @@
...
+	hsFrame, err = readHSFrame(stream)
+	if err != nil {
+		return nil, errors.Wrap(err, "error reading HS frame")
+	}
+
+	if hsFrame.FrameType() != HSFrameTypeDMSGAccept {
+		return nil, ErrWrongHSFrameTypeReceived
+	}
+
+	return stream, nil
```
evanlinjin: We cannot just return `stream` here. We need a structure that knows local/remote dmsg pk and port.

pkg/app2/hsframe.go
```diff
@@ -0,0 +1,143 @@
...
+	HSFrameTypeLen   = 1
+	HSFramePKLen     = 33
+	HSFramePortLen   = 2
+)
+
+// HSFrameType identifies the type of a handshake frame.
+type HSFrameType byte
+
+const (
+	HSFrameTypeDMSGListen HSFrameType = 10 + iota
```
evanlinjin: I prefer the following naming convention: `DmsgListenType`, `DmsgListeningType`, `DmsgAcceptType`, etc.

It is a bit less wordy.

pkg/app2/hsframe.go
```diff
@@ -0,0 +1,143 @@
...
+func newHSFrame(procID ProcID, frameType HSFrameType, bodyLen int) HSFrame {
+	hsFrame := make(HSFrame, HSFrameHeaderLen+bodyLen)
+
+	hsFrame.SetProcID(procID)
+	hsFrame.SetFrameType(frameType)
+
+	return hsFrame
+}
+
+func NewHSFrameDMSGListen(procID ProcID, local routing.Addr) HSFrame {
```
evanlinjin: `NewDmsgListenFrame` would be better. Please follow this convention for the functions below too.

pkg/app2/hsframe.go
```diff
@@ -0,0 +1,143 @@
...
+	return hsFrame
+}
+
+// ProcID gets ProcID from the HSFrame.
+func (f HSFrame) ProcID() ProcID {
+	return ProcID(binary.BigEndian.Uint16(f))
+}
+
+// SetProcID sets ProcID for the HSFrame.
+func (f HSFrame) SetProcID(procID ProcID) {
```
evanlinjin: Not sure if the `Set` member functions for `HSFrame` will be useful. Frames are created as is and is read by the destination.

pkg/app2/listener.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	listenerBufSize = 1000
+)
+
+var (
+	ErrListenerClosed = errors.New("listener closed")
+)
+
+type Listener struct {
```
evanlinjin, line 13: Listener needs to handle `Close` in a thread-safe manner. Please use [`snet.Listener`](https://github.com/skycoin/skywire/blob/6439c70ee1ca10d2086c3721fe364800c27c00aa/pkg/snet/stcp/client.go#L56) as an example.

In fact, maybe the entire file will be helpful as it is similar to what we are doing here.

pkg/app2/client.go
evanlinjin: It might help to group *pointer inputs* in a structure if we end up with too many inputs (similar to how stdlib `rpc` is done).

pkg/app2/server.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		s.appsMx.Lock()
+		if _, ok := s.apps[conn.RemoteAddr().String()]; ok {
+			s.logger.WithError(ErrPortAlreadyBound).Error("error storing session")
+		}
+
+		session, err := yamux.Server(conn, nil)
+		if err != nil {
+			return errors.Wrap(err, "error creating yamux session")
```
evanlinjin: You forgot a `s.appsMx.Unlock()` here.

evanlinjin (changes requested):
Good start.

pkg/app2/listener.go
```diff
@@ -41,6 +53,15 @@ func (l *Listener) Accept() (net.Conn, error) {
 		return nil, ErrListenerClosed
 	}
 
+	hsFrame := NewHSFrameDMSGAccept(l.procID, routing.Loop{
```
evanlinjin: Have you decided to follow what is proposed here: https://discordapp.com/channels/466450852098605059/616123208357380106/618822964917567498

Or have you decided to do something else? If so, can you define the new communication protocol and why you have chosen this route?

pkg/app2/server.go
```diff
@@ -110,68 +116,151 @@ func (s *Server) serveClient(session *yamux.Session) error {
...
+
+						respHSFrame = NewHSFrameDMSGListening(hsFrame.ProcID(), routing.Addr{
+							PubKey: pk,
+							Port:   routing.Port(port),
+						})
+					}
+				}
 			}
+		case HSFrameTypeDMSGDial:
+			localPort := binary.BigEndian.Uint16(hsFrame[HSFrameHeaderLen+HSFramePKLen:])
```
evanlinjin: I hope this is WIP and the logic here will be extracted to it's own functions?

pkg/app2/client.go
Darkren: Done

pkg/app2/client.go
Darkren: Done

pkg/app2/server_rpc.go
```diff
@@ -0,0 +1,244 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/skycoin/dmsg"
+	"github.com/skycoin/skycoin/src/util/logging"
+
+	"github.com/skycoin/skywire/pkg/routing"
+)
+
+// ServerRPC is a RPC interface for the app server.
+type ServerRPC struct {
```
evanlinjin: Rename to `RPCGateway`.

pkg/app2/server_rpc_client.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"net/rpc"
+
+	"github.com/skycoin/skywire/pkg/routing"
+)
+
+//go:generate mockery -name ServerRPCClient -case underscore -inpkg
+
+// ServerRPCClient describes RPC interface to communicate with the server.
+type ServerRPCClient interface {
```
evanlinjin: Rename to `RPCClient`.

pkg/app2/client.go
```diff
@@ -0,0 +1,87 @@
...
+	rpc    ServerRPCClient
+	porter *netutil.Porter
+}
+
+// NewClient creates a new `Client`. The `Client` needs to be provided with:
+// - log: Logger instance.
+// - localPK: The local public key of the parent skywire visor.
+// - pid: The procID assigned for the process that Client is being used by.
+// - rpc: RPC client to communicate with the server.
+func NewClient(localPK cipher.PubKey, pid ProcID, rpc ServerRPCClient, porter *netutil.Porter) *Client {
```
evanlinjin: `porter` does not need to be an input. It can be created here. The `minEph` value will never change.

pkg/app2/client.go
```diff
@@ -0,0 +1,87 @@
...
+	return &Client{
+		pk:     localPK,
+		pid:    pid,
+		rpc:    rpc,
+		porter: porter,
+	}
+}
+
+// Dial dials the remote node using `remote`.
+func (c *Client) Dial(remote routing.Addr) (net.Conn, error) {
```
evanlinjin: Rename to `DmsgDial`

pkg/app2/client.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+		remote:        remote,
+		freeLocalPort: free,
+	}
+
+	return conn, nil
+}
+
+// Listen listens on the specified `port` for the incoming connections.
+func (c *Client) Listen(port routing.Port) (net.Listener, error) {
```
evanlinjin: `DmsgListen`

pkg/app2/client.go
evanlinjin: use `remote dmsg.Addr` as we are dialling via `dmsg`.

pkg/app2/conn.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"net"
+	"time"
+
+	"github.com/skycoin/skywire/pkg/routing"
+)
+
+// Conn is a connection from app client to the server.
+// Implements `net.Conn`.
+type Conn struct {
```
evanlinjin, line 12: for fields `local` and `remote`, have them of `net.Addr` type. That way we can put `dmsg.Addr` or `routing.Addr` (depending on the connection type).

pkg/app2/listener.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"net"
+
+	"github.com/skycoin/skywire/pkg/routing"
+)
+
+// Listener is a listener for app server connections.
+// Implements `net.Listener`.
+type Listener struct {
```
evanlinjin, line 13: Maybe rename this to `DmsgListener` or find a way to make it work with all types of `Listener`s (I hope you know what I mean).

pkg/app2/manager.go
```diff
@@ -0,0 +1,101 @@
...
+	m.values[nxtKey] = nil
+	m.lstKey = nxtKey
+
+	m.mx.Unlock()
+	return &nxtKey, nil
+}
+
+// getAndRemove removes value specified by `key` from the manager instance and
+// returns it.
+func (m *manager) getAndRemove(key uint16) (interface{}, error) {
```
evanlinjin: Maybe `pop` will be a better name?

pkg/app2/server_rpc.go
```diff
@@ -0,0 +1,244 @@
...
+	return &ServerRPC{
+		dmsgC: dmsgC,
+		lm:    newManager(),
+		cm:    newManager(),
+		log:   log,
+	}
+}
+
+// Dial dials to the remote.
+func (r *ServerRPC) Dial(remote *routing.Addr, connID *uint16) error {
```
evanlinjin: `DmsgDial` (unless we can make this dial to all sorts).

pkg/app2/server_rpc.go
```diff
@@ -0,0 +1,244 @@
...
+
+	if err := r.cm.set(*connID, tp); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Listen starts listening.
+func (r *ServerRPC) Listen(local *routing.Addr, lisID *uint16) error {
```
evanlinjin: `DmsgListen`

pkg/app2/server_rpc.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// AcceptResp contains response parameters for `Accept`.
+type AcceptResp struct {
+	Remote routing.Addr
+	ConnID uint16
+}
+
+// Accept accepts connection from the listener specified by `lisID`.
+func (r *ServerRPC) Accept(lisID *uint16, resp *AcceptResp) error {
```
evanlinjin: `DmsgAccept`? Question mark because maybe we can make all these *networks* use the same function.

evanlinjin (changes requested):
Beautifully written code.

Although there is some implementation queries. Specifically, we will need to support multiple *networks* in the future:
- `dmsg`
- `skywire routes`

My comments are mostly regarding this. Hence, you do not need to follow the exact comments, but please take some time to think of an optimal solution.

pkg/app2/server_rpc.go
Darkren: Done

pkg/app2/server_rpc_client.go
Darkren: Done

pkg/app2/client.go
Darkren: Done

pkg/app2/manager.go
Darkren: Done

pkg/app2/rpc_client.go
```diff
@@ -0,0 +1,97 @@
...
+
+// NewRPCClient constructs new `rpcClient`.
+func NewRPCClient(rpc *rpc.Client) RPCClient {
+	return &rpcCLient{
+		rpc: rpc,
+	}
+}
+
+// Dial sends `Dial` command to the server.
+func (c *rpcCLient) Dial(remote network.Addr) (uint16, error) {
```
evanlinjin: Sorry should have picked this out earlier, but we need to use the ephemeral local address of the actual connection. This should return the local address as well as the identifier.

pkg/app2/rpc_client.go
```diff
@@ -0,0 +1,97 @@
...
+	var lisID uint16
+	if err := c.rpc.Call("Listen", &local, &lisID); err != nil {
+		return 0, err
+	}
+
+	return lisID, nil
+}
+
+// Accept sends `Accept` command to the server.
+func (c *rpcCLient) Accept(lisID uint16) (uint16, network.Addr, error) {
```
evanlinjin: We need to name the output variables or have a comment that makes it clear that it is the remote address that we are returning here.

pkg/app2/rpc_client.go
```diff
@@ -0,0 +1,97 @@
...
+		return n, err
+	}
+
+	return n, nil
+}
+
+// Read sends `Read` command to the server.
+func (c *rpcCLient) Read(connID uint16, b []byte) (int, []byte, error) {
+	var resp ReadResp
+	if err := c.rpc.Call("Read", &connID, &resp); err != nil {
```
evanlinjin: `(net.Conn).Read(b []byte) (int, error)` relies heavily on the length of the slice `b`. We need to pass this information to the RPC gateway. The request for this RPC call should be of a structure similar to the following:
```golang
type ReadReq struct {
    ConnID uint16
    BLen int // Length of input `b` of `(net.Conn).Read`.
}
```

pkg/app2/client.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/skycoin/skywire/pkg/app2/network"
+	"github.com/skycoin/skywire/pkg/routing"
+)
+
+// Client is used by skywire apps.
+type Client struct {
+	pk     cipher.PubKey
+	pid    ProcID
+	rpc    RPCClient
+	porter *netutil.Porter
```
evanlinjin: Sorry for the changes, but `netutil.Porter` is not the best structure to use here. The local ephemeral keys should be determined by the `network` that we are using, not by `app2.Client`.

Additionally, we need to close all associated `net.Conn`s and `net.Listeners` of this `app2.Client` when it closes. We should keep track of all `connID`s and `lisID`s here and have a `Client.Close` method.

pkg/app2/listener.go
```diff
@@ -0,0 +1,43 @@
...
+		id:     connID,
+		rpc:    l.rpc,
+		local:  l.addr,
+		remote: remote,
+	}
+
+	return conn, nil
+}
+
+// TODO: should unblock all called `Accept`s with errors
```
evanlinjin: This should happen already right? Unless if there is a bug in `dmsg`.

pkg/app2/manager.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+var (
+	errNoMoreAvailableValues = errors.New("no more available values")
+)
+
+// manager manages allows to store and retrieve arbitrary values
+// associated with the `uint16` key in a thread-safe manner.
+// Provides method to generate key.
+type manager struct {
```
evanlinjin: Maybe `idManager` will be a better name?

pkg/app2/network/dmsg_conn.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"net"
+	"time"
+
+	"github.com/skycoin/skywire/pkg/routing"
+
+	"github.com/skycoin/dmsg"
+)
+
+type DMSGConn struct {
```
evanlinjin: Why is this needed? Can't we just use `*dmsg.Transport` directly?

pkg/app2/rpc_gateway.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/pkg/errors"
+	"github.com/skycoin/skycoin/src/util/logging"
+
+	"github.com/skycoin/skywire/pkg/app2/network"
+)
+
+// RPCGateway is a RPC interface for the app server.
+type RPCGateway struct {
+	lm  *manager
+	cm  *manager
```
evanlinjin: Let's add some brief comments explaining the fields `lm` and `cm`.

evanlinjin (changes requested):
Clean code and good work so far. Please review my comments.

pkg/app2/network/dmsg_conn.go
Darkren: Not really. I used it directly, but since we have a generic `network.Addr` which we return from the server, we also need conns and listeners to return this type underneath the `net.Addr` interface. So, I just had to do this

pkg/app2/listener.go
Darkren: I'll check it

pkg/app2/rpc_client.go
Darkren: I think that I'd better put it into the `Client.Dial`, since it constructs the actual connection. This way, we'll let `rpcClient` do it's rpc stuff without bothering it with the additional work. I think it's more of a client's responsibility to assign port. What do you think?

pkg/app2/client.go
Darkren: Hm, but this way, I'll need to do my own porter implementation, right?

pkg/app2/listener.go
Darkren: No, `Accept` blocks on the RPC `Accept` call. If we close listener, RPC call doesn't unblock, and we can't just close RPC here. Some smart move needs to be here

pkg/app2/network/dmsg_conn.go
Darkren: But it actually messes up some of my logic. Need to think a bit

pkg/app2/rpc_client.go
evanlinjin: > This way, we'll let rpcClient do it's rpc stuff without bothering it with the additional work.

What additional work are you referring to here?

> I think it's more of a client's responsibility to assign port.

I'm not sure if I'm understanding you correctly here. But the main idea of the `app2` library is to perform *network* operations on the visor node. Hence, we want to emulate as much of this interaction as possible, so the entity that uses `app2` sees closely to what they would see if they used `dmsg` or `router` directly. Hence, the both the local and remote addresses of the `dmsg.Transport` between `app2` and `visor`, needs to look the same as between `visor` and the remote `visor`.

Hope that makes sense.

pkg/app2/client.go
evanlinjin: No, the ephemeral port is decided by the network. At most, we need one or two maps.

pkg/app2/listener.go
evanlinjin: Given the definition of [`net.Listener`](https://golang.org/pkg/net/#Listener), the following line should unblock when we call `.Close` on the same listener:

https://github.com/skycoin/skywire/blob/e243192a1e1beccb0fcc3658d9ac6a4cb5243094/pkg/app2/rpc_gateway.go#L97

If it doesn't, it means there is a bug with `dmsg`.

evanlinjin: Is there somewhere else where the `(*RPCGateway).Accept` blocks?

pkg/app2/network/dmsg_conn.go
evanlinjin: The only two things differentiating `dmsg.Transport` and `DMSGConn` is the local and remote addresses. Won't it be better to create a universal structure that can be used with all `net.Conn`s?

For example:

```golang
type WrappedConn struct {
    net.Conn
    lAddr Addr
    rAddr Addr
}

func (c WrappedConn).LocalAddr() net.Addr { return c.lAddr }
func (c WrappedConn).RemoteAddr() net.Conn { return c.rAddr }
```

pkg/app2/rpc_client.go
Darkren: @evanlinjin The additional work is to assign port. `rpcClient`'s only responsibility is to handle RPC calls. The result of the call is being returned to the `Client` which interprets it some way. `Client` constructs `Conn` in my implementation, and that's where the port should be assigned imo

pkg/app2/client.go
Darkren: @evanlinjin I understand your intention, but I reserve ports through porter. If I remove porter, then I'll need some replacement for it

pkg/app2/listener.go
Darkren: @evanlinjin yeah, my mistake, everything should be fine

pkg/app2/network/dmsg_conn.go
Darkren: @evanlinjin the problem here is the specific implementation of `net.Conn`. `DMSGConn` wraps transport, so it can assert `dmsg.Addr` type to `net.Addr` and then safely convert `dmsg.Addr` to `network.Addr` also assigning valid network type. If we wrap it the way you suggest, we won't be able to operate on the underlying addr struct. Well, we can if we do a switch over all possible types which can be there. If there's what you'd like to see here, I'll do it this way

pkg/app2/rpc_client.go
evanlinjin: We do not need to assign port. When we call `dmsg.Dial`, we get a local ephemeral port. We should be using this for the associated `app2.Conn`.

```golang
var client *dmsg.Client
conn, err := client.Dial()
if err != nil { panic(err) }
lAddr := conn.LocalAddr()
lPort := lAddr.(dmsg.Addr).Port // Get local ephemeral port here.
```

pkg/app2/client.go
evanlinjin: We do not need to keep track of ports here. We just need to keep track of `connID`s and `lisID`s so that these can be closed when `app2.Client` closes.

pkg/app2/network/dmsg_conn.go
evanlinjin: > Well, we can if we do a switch over all possible types which can be there. If there's what you'd like to see here, I'll do it this way

Yes please.

Darkren: @evanlinjin nailed it, please check it out

pkg/app2/rpc_client.go
Darkren: @evanlinjin Done

pkg/app2/client.go
Darkren: @evanlinjin and this one is done. Just need to adjust tests after all modifications

pkg/app2/network/addr.go
```diff
@@ -0,0 +1,51 @@
...
+func (a Addr) String() string {
+	if a.Port == 0 {
+		return fmt.Sprintf("%s:~", a.PubKey)
+	}
+	return fmt.Sprintf("%s:%d", a.PubKey, a.Port)
+}
+
+// WrapAddr asserts type of the passed `net.Addr` and converts it
+// to `Addr` if possible.
+func WrapAddr(addr net.Addr) (Addr, error) {
```
evanlinjin: Maybe `ConvertAddr` will be a better name.

pkg/app2/network/addr.go
```diff
@@ -0,0 +1,51 @@
...
+	switch a := addr.(type) {
+	case dmsg.Addr:
+		return Addr{
+			Net:    TypeDMSG,
+			PubKey: a.PK,
+			Port:   routing.Port(a.Port),
+		}, nil
+	default:
+		return Addr{}, errUnknownAddrType
```
evanlinjin: Instead of returning an error, we can panic here to make things easier.

pkg/app2/rpc_client.go
```diff
@@ -0,0 +1,105 @@
...
+
+// NewRPCClient constructs new `rpcClient`.
+func NewRPCClient(rpc *rpc.Client) RPCClient {
+	return &rpcCLient{
+		rpc: rpc,
+	}
+}
+
+// Dial sends `Dial` command to the server.
+func (c *rpcCLient) Dial(remote network.Addr) (connID uint16, assignedPort routing.Port, err error) {
```
evanlinjin: rename `assignedPort` to `localPort`.

pkg/app2/client.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	conn.freeConn = free
+
+	return conn, nil
+}
+
+// Listen listens on the specified `port` for the incoming connections.
+func (c *Client) Listen(n network.Type, port routing.Port) (net.Listener, error) {
+	ok, freePort := c.porter.Reserve(uint16(port), nil)
```
evanlinjin: We don't need to reserve anything here. We should rely on the *network*.

pkg/app2/client.go
evanlinjin: We should get rid of `porter` entirely.

evanlinjin (changes requested):
Good stuff. Nearly there.

Issue #550, opened 2019-09-04
Data races in hypervisor


nkryuchkov:
Commits:
- `skywire`: `2a81b33e`
- `skywire-services`: `08d6bf47`
- `dmsg`: `b781e3cb`

How to reproduce:
1. Run the generic test environment
2. Send a `GET` request to `localhost:8080/api/nodes` (other endpoints seem affected too)

Output:

1. 
```
WARNING: DATA RACE
Read at 0x00c00000fe50 by goroutine 21:
  github.com/skycoin/dmsg/noise.(*Conn).Write()
      /Users/nkryuchkov/dmsg/noise/net.go:178 +0xac
  bufio.(*Writer).Flush()
      /usr/local/Cellar/go/1.12.9/libexec/src/bufio/bufio.go:590 +0x13f
  net/rpc.(*gobClientCodec).WriteRequest()
      /usr/local/Cellar/go/1.12.9/libexec/src/net/rpc/client.go:224 +0x116
  net/rpc.(*Client).send()
      /usr/local/Cellar/go/1.12.9/libexec/src/net/rpc/client.go:91 +0x38e
  net/rpc.(*Client).Go()
      /usr/local/Cellar/go/1.12.9/libexec/src/net/rpc/client.go:316 +0x153
  github.com/skycoin/skywire/pkg/visor.(*rpcClient).Call()
      /usr/local/Cellar/go/1.12.9/libexec/src/net/rpc/client.go:322 +0x153
  github.com/skycoin/skywire/pkg/visor.(*rpcClient).Summary()
      /Users/nkryuchkov/skywire/pkg/visor/rpc_client.go:65 +0xa7
  github.com/skycoin/skywire/pkg/hypervisor.(*Node).getNodes.func1()
      /Users/nkryuchkov/skywire/pkg/hypervisor/hypervisor.go:189 +0x1fa
  net/http.HandlerFunc.ServeHTTP()
...

Previous write at 0x00c00000fe50 by goroutine 11:
  github.com/skycoin/dmsg/noise.(*Listener).Accept()
      /Users/nkryuchkov/dmsg/noise/net.go:233 +0x6ec
  github.com/skycoin/skywire/pkg/hypervisor.(*Node).ServeRPC()
      /Users/nkryuchkov/skywire/pkg/hypervisor/hypervisor.go:64 +0x272
  github.com/skycoin/skywire/cmd/hypervisor/commands.glob..func3.1()
[...]

PR #554, opened 2019-09-05
proper mux lock

ivcosla:
Fixed data races at #550


nkryuchkov (approved):
I've run the tests and seen no data races at all. Good job!

PR #562, opened 2019-09-06
[WIP] [M2] Implement router2 according to specs

### LINKED ISSUE #545: [M2] Implement router2 according to specs
The goal is to split out and define the responsiblities within `router` to allow for easier integration of future features, as well as make the codebase easier to understand.

**Specifications:** https://github.com/watercompany/skywire-specs/blob/master/proposals/router2.md

nkryuchkov:
Closes #545 

### `routing.Rule`:
- [x] Change `routing.Rule` structure

### `routing.Table`:
- [x] Remove BoltDB and managed route tables
- [x] Change `route.Table` interface
- [x] Remove old methods from `route.Table` 

### Setup node:
- [x] Create `Client` type
- [x] Remove `Close~` packet types
- [x] Change `Hop` structure
- [x] Change `Route` structure
- [x] Use `Desc` and `KeepAlive` fields of `Route`
- [x] Use one connection per visor, per request
- [x] Use failure codes
- [ ] Fix tests

### Router:
- [ ] Rewrite
- [ ] Keepalive
- [ ] Fix tests

### Specs:
- [ ] Update


pkg/routing/rule.go
```diff
@@ -14,207 +13,421 @@ import (
...
+	r.assertLen(RuleHeaderSize)
 	return RuleType(r[8])
 }
 
-// RouteID returns RouteID from the rule: reverse ID for an app rule
-// and next ID for a forward rule.
-func (r Rule) RouteID() RouteID {
-	return RouteID(binary.BigEndian.Uint32(r[9:]))
+// SetType sets type of a rule.
+func (r Rule) SetType(t RuleType) {
```
evanlinjin: Remove the `Set...` member functions. Rules are only created as a whole and should not be modified.

evanlinjin (changes requested):

pkg/router/route_manager.go
```diff
@@ -6,9 +6,7 @@ import (
 	"errors"
```
evanlinjin: I rather get rid of this file completely. The logic here should be moved to `router.Router` and `setup.Client`.

Maybe first ensure the work in `setup` and route finder is done before work on `router`.

pkg/routing/loop.go
```diff
@@ -28,20 +28,20 @@ type LoopDescriptor struct {
```
evanlinjin: In the new implementation, there will be no such thing as a `Loop`.

pkg/routing/table.go
```diff
@@ -0,0 +1,212 @@
...
+
+// DefaultConfig represents the default configuration of routing table.
+func DefaultConfig() Config {
+	return Config{
+		GCInterval: DefaultGCInterval,
+	}
+}
+
+// New returns an in-memory routing table implementation.
+func New() Table {
```
evanlinjin: We don't need this. Please remove.

pkg/routing/table.go
```diff
@@ -0,0 +1,212 @@
...
+	return NewWithConfig(DefaultConfig())
+}
+
+// NewWithConfig returns an in-memory routing table implementation with a specified configuration.
+func NewWithConfig(config Config) Table {
```
evanlinjin: `NewTable`

pkg/routing/table.go
```diff
@@ -0,0 +1,212 @@
...
+		if !mt.ruleIsTimedOut(k, v) && v.RouteDescriptor() == desc {
+			rules = append(rules, v)
+		}
+	}
+
+	return rules
+}
+
+// RuleEntry is a pair of a RouteID and a Rule.
+type RuleEntry struct {
```
evanlinjin: We don't need this. `Rule` is sufficient.

`Rule` contains both the `RouteID` key and the next route ID (if `ForwardRule` or `IntermediaryForwardRule`).

pkg/routing/table.go
```diff
@@ -0,0 +1,212 @@
...
+				Rule:    v,
+			}
+			rules = append(rules, entry)
+		}
+	}
+
+	return rules
+}
+
+func (mt *memTable) DelRules(routeIDs []RouteID) {
```
evanlinjin: `keys []RouteID`

pkg/routing/table.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, routeID := range routeIDs {
+		mt.Lock()
+		mt.delRule(routeID)
+		mt.Unlock()
+	}
+}
+
+func (mt *memTable) delRule(routeID RouteID) {
```
evanlinjin: `key RouteID`

pkg/routing/table.go
```diff
@@ -0,0 +1,212 @@
...
+	mt.Lock()
+	defer mt.Unlock()
+
+	mt.rules[routeID] = rule
+	mt.activity[routeID] = time.Now()
+
+	return nil
+}
+
+func (mt *memTable) Rule(routeID RouteID) (Rule, error) {
```
evanlinjin: `key RouteID`

pkg/routing/table.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if mt.nextID == math.MaxUint32 {
+		return 0, ErrNoAvailableRoutes
+	}
+
+	mt.nextID++
+	return mt.nextID, nil
+}
+
+func (mt *memTable) SaveRule(routeID RouteID, rule Rule) error {
```
evanlinjin: `SaveRule(r Rule) error`

pkg/routing/table.go
```diff
@@ -0,0 +1,212 @@
...
+		rules:    map[RouteID]Rule{},
+		activity: make(map[RouteID]time.Time),
+	}
+
+	go mt.gcLoop()
+
+	return mt
+}
+
+func (mt *memTable) ReserveKey() (RouteID, error) {
```
evanlinjin: `ReserveKey() (key RouteID, err error)`

evanlinjin (changes requested):

evanlinjin:
Good work so far.

Looking at the codebase, it seems that your current approach is to make incremental changes and ensure those changes work with the current codebase.

However, this approach is time consuming as we are doing mostly a rewrite.

Please consider this as a rewrite rather than a refactor.

I recommend prioritising the following:
- Understand the end goal as specified within the specs.
- Ensure the individual components within the specs work.

Later on, once we are sure the components work as expected, we can thing about replacing the current codebase.

P.S. Please fix merge conflicts.

pkg/router/router.go
```diff
@@ -3,32 +3,30 @@ package router
...
-
 	"github.com/skycoin/dmsg/cipher"
 	"github.com/skycoin/skycoin/src/util/logging"
 
 	"github.com/skycoin/skywire/pkg/app"
-	routeFinder "github.com/skycoin/skywire/pkg/route-finder/client"
+	rfclient "github.com/skycoin/skywire/pkg/route-finder/client"
 	"github.com/skycoin/skywire/pkg/routing"
 	"github.com/skycoin/skywire/pkg/setup"
+	rsclient "github.com/skycoin/skywire/pkg/setup/client"
```
evanlinjin: Rename the actual package name to be `"github.com/skycoin/skywire/pkg/route-finder/rsclient"`

pkg/router/router.go
```diff
@@ -3,32 +3,30 @@ package router
...
 	"time"
 
-	"github.com/skycoin/skywire/pkg/snet"
-
 	"github.com/skycoin/dmsg/cipher"
 	"github.com/skycoin/skycoin/src/util/logging"
 
 	"github.com/skycoin/skywire/pkg/app"
-	routeFinder "github.com/skycoin/skywire/pkg/route-finder/client"
+	rfclient "github.com/skycoin/skywire/pkg/route-finder/client"
```
evanlinjin: Rename the actual package name to be `"github.com/skycoin/skywire/pkg/routefinder/rfclient"`

evanlinjin: And `github.com/skycoin/skywire/pkg/route-finder` should be `github.com/skycoin/skywire/pkg/routefinder`

pkg/router/router.go
```diff
@@ -69,36 +63,42 @@ type Router struct {
...
+	sl *snet.Listener
+	rt routing.Table
+
+	rsc rsclient.Client
 
 	wg sync.WaitGroup
 	mx sync.Mutex
+
+	OnConfirmLoop func(loop routing.Loop, rule routing.Rule) (err error)
+	OnLoopClosed  func(loop routing.Loop) error
```
evanlinjin: Why do we need the `OnConfirmLoop` and `OnLoopClosed` members? Can we not call `r.confirmLoop` and `r.loopClosed` directly?

pkg/router/router.go
```diff
@@ -69,36 +63,42 @@ type Router struct {
...
-		SetupPKs:               config.SetupNodes,
-		GarbageCollectDuration: config.GarbageCollectDuration,
-		OnConfirmLoop:          r.confirmLoop,
-		OnLoopClosed:           r.loopClosed,
-	})
-	if err != nil {
-		return nil, err
-	}
-	r.rm = rm
+	r.rsc = rsclient.New(n, sl, config.SetupNodes, r.handleSetupConn)
```
evanlinjin: Callbacks are confusing. It will be better to create a new structure named `SetupGateway` which has `Router` as a member. Hence, `SetupGateway` will have member function: `SetupGateway.handleSetupConn` and it's own `SetupGateway.Serve`, removing the need for callbacks.

pkg/setup/client/client.go
```diff
@@ -0,0 +1,97 @@
...
+	"net"
+
+	"github.com/skycoin/dmsg/cipher"
+	"github.com/skycoin/skycoin/src/util/logging"
+
+	"github.com/skycoin/skywire/pkg/snet"
+)
+
+// Client interacts with setup nodes.
+type Client interface {
```
evanlinjin: It seems as though I was wrong with having a `setup.Client` able to clean up the logic. Let's try the `router.SetupGateway` structure instead (unless you have a better idea?).

pkg/setup/node.go
```diff
@@ -7,16 +7,14 @@ import (
...
 	"github.com/skycoin/dmsg/cipher"
 	"github.com/skycoin/dmsg/disc"
 	"github.com/skycoin/skycoin/src/util/logging"
 
 	"github.com/skycoin/skywire/pkg/metrics"
 	"github.com/skycoin/skywire/pkg/routing"
+	"github.com/skycoin/skywire/pkg/snet"
 )
 
 // Node performs routes setup operations over messaging channel.
```
evanlinjin, line 19: `setup.Node` still needs to follow spec: https://github.com/watercompany/skywire-specs/blob/master/proposals/router2.md#setup-node-changes

evanlinjin: Sorry for the change of mind, but instead of defining our own communication protocol for setup, we should use stdlib's `rpc`.

### Instructions

We need two *gateway* structures (which we use to call `(*rpc.Server).Register` with):
- One hosted on `router` that handles requests from `setup`.
- One hosted on `setup` that handles requests from `router`.

So...
- Instead of *Packet Types*, we use member functions for the *gateway* structures.
- Instead of *Packet Body*, we use the inputs of the member functions of the *gateway* structures.

### Examples

Here are some examples for inspiration:
- `app2` implementation: https://github.com/Darkren/skywire/tree/feature/app2/pkg/app2
- `dmsgexec`: https://github.com/SkycoinProject/dmsgexec/blob/master/rpc.go

evanlinjin (changes requested):
### Summary of review

Good work so far. I really like what you have done for the routing rules and routing table.

Please ensure setup node is up to spec. Additionally, I think we should use `rpc` instead of defining our own communication protocol for it. Apologies for the change of mind.

I've also mentioned in this review that `router` should be split into `Router` and `SetupGateway`. Where `SetupGateway` has `Router` as a member.

### Additional Comments

Also ensure that the route finder changes are integrated (please talk to @ivcosla ).

Also note that there is no such thing as `Loop` any more.

pkg/router/router.go
evanlinjin: My point is that the organisation of our structures should be like a tree instead of cyclic. That way, things are easier to understand.

In some cases, callbacks are great, just not here as it makes us read the code in a *cyclic* manner.

pkg/setup/client/client.go
aussieboy88: Client

pkg/router/router.go
nkryuchkov: @evanlinjin I left them for the compatibility with tests. They were modified in tests. However, they will be removed in the next commits.

evanlinjin: Sounds good.

evanlinjin:
How's progress?

nkryuchkov:
@evanlinjin I've just implemented the usage of `net/rpc` instead of own protocol. Integrating the changes with `RouteGroup`'s right now.

pkg/app/app.go
```diff
@@ -38,16 +39,16 @@ type Config struct {
...
 
-// App represents client side in app's client-server communication
+// App represents rsclient side in app's rsclient-server communication
 // interface.
 type App struct {
 	config Config
 	proto  *Protocol
 
-	acceptChan chan [2]routing.Addr
+	acceptChan chan []routing.Route
```
evanlinjin: Would it be better to use `routing.RouteDescriptor` here?

pkg/app/packet.go
```diff
@@ -4,6 +4,6 @@ import "github.com/skycoin/skywire/pkg/routing"
 
 // Packet represents message exchanged between App and Node.
 type Packet struct {
-	Loop    routing.Loop `json:"loop"`
-	Payload []byte       `json:"payload"`
+	Route   routing.Route `json:"route"`
```
evanlinjin: Would `routing.RouteDescriptor` here be better?

pkg/router/rpc_server.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/skycoin/skywire/pkg/app"
+	"github.com/skycoin/skywire/pkg/routing"
+	"github.com/skycoin/skywire/pkg/setup"
+)
+
+func (r *Router) AddRules(rules []routing.Rule, routeIDs *[]routing.RouteID) error {
```
evanlinjin: We should create two structures:
- `Router`
- `RouterGateway`

Where `RouterGateway` has rpc-specific these members, and uses `Router`.

evanlinjin: We should also create a `RouterClient` structure,

pkg/setup/idreservoir.go
```diff
@@ -97,60 +100,46 @@ func (rm RulesMap) String() string {
...
-// - firstRID: the first visor's route ID.
-// - lastRID: the last visor's route ID (note that there is no rule set for this ID yet).
-// - err: an error (if any).
-func SaveForwardRules(rules RulesMap, idc *idReservoir, keepAlive time.Duration, route routing.Route) (firstRID, lastRID routing.RouteID, err error) {
+// - a map that relates a slice of routing rules to a given visor's public key.
+// - an error (if any).
+func (idr *idReservoir) GenerateRules(route routing.Route) (RulesMap, error) {
+	rules := make(RulesMap)
 
 	// 'firstRID' is the first visor's key routeID - this is to be returned.
```
evanlinjin: Remember to update the comments.

pkg/setup/node.go
```diff
@@ -56,220 +51,42 @@ func NewNode(conf *Config, metrics metrics.Recorder) (*Node, error) {
...
-	}
-
-	return NewSetupProtocol(tr), nil
-}
 
-func (sn *Node) closeProto(proto *Protocol) {
-	if err := proto.Close(); err != nil {
-		sn.Logger.Warn(err)
+		sn.logger.WithField("requester", conn.RemotePK()).Infof("Received request.")
+		go sn.rpcSrv.ServeConn(conn.Conn)
```
evanlinjin: We should use a separate `setup.Gateway` structure. This will serve two purposes:

1. We can completely separate setup logic, and the code that only serves as a *gateway* between the setup node and visor node.

2. We can verify requests.

For example, if visor A wants to `CreateRoutes`, we need to verify that the route requested needs to start from A. We can use `setup.Gateway` to store the *context* of the request if we create a new `setup.Gateway` for each request:

```golang
package setup

type Gateway struct {
    reqPK cipher.PubKey // requester's public key
    sn *Node // setup node
}

func (sn *Node) Serve() error {
    for {
        conn, err := sn.dmsgL.AcceptTransport()
        if err != nil { return err }
        rpcS := rpc.NewServer()
        rpcS.RegisterName(&Gateway{reqPK: conn.RemotePK(), sn:sn})
        go rpcS.ServeConn(conn)
    }
}
```

pkg/setup/visor_rpc_helpers.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/skycoin/dmsg/cipher"
+
+	"github.com/skycoin/skywire/pkg/routing"
+)
+
+func (sn *Node) visorAddRules(ctx context.Context, pk cipher.PubKey, rules []routing.Rule) ([]routing.RouteID, error) {
```
evanlinjin: These functions do not need to be members of `setup.Node`. We can just have `*dmsg.Client` as an additional input.

evanlinjin (changes requested):
Good work so far. Please check my comments.

pkg/app/app.go
nkryuchkov: @evanlinjin It would be, thanks

pkg/app/packet.go
nkryuchkov: @evanlinjin Indeed. Thanks.

pkg/router/rpc_server.go
nkryuchkov: @evanlinjin Good idea

pkg/setup/visor_rpc_helpers.go
nkryuchkov: @evanlinjin 
I made them members of `setup.Node` to use its logger when the closing of router RPC client fails. Creating its own logger for the RPC client didn't seem necessary for me. Anyway, those methods get called from `setup.Node` methods only. 

Do you think it would be better to create its own logger?

pkg/setup/node.go
nkryuchkov: @evanlinjin Good suggestion.

pkg/setup/visor_rpc_helpers.go
evanlinjin: > Do you think it would be better to create its own logger?

No, that would be a little extreme.

Good point, I did not see this. How about having an additional `log logrus.StdLogger` as input?

nkryuchkov: @evanlinjin Actually, we can just remove the logging dependency in those functions and just return an error if the closing of client fails. Then we can get rid of `Node` dependency. What do you think?

evanlinjin: That can work too. But failing to close shouldn't result in route setup failure later on.

nkryuchkov: @evanlinjin I'll think how to do this in the best way. An additional logging argument would be helpful, however, I don't like this signature:
```golang
func visorAddRules(ctx context.Context, log logrus.StdLogger, dmsgC *dmsg.Client, pk cipher.PubKey, rules []routing.Rule) ([]routing.RouteID, error)
```

evanlinjin: @nkryuchkov Personally I think it looks okay :sweat_smile: 

Although slightly lengthy, the inputs are easy to understand.

nkryuchkov: @evanlinjin I'll do it this way then

pkg/app/app.go
```diff
@@ -44,10 +44,10 @@ type App struct {
 	config Config
 	proto  *Protocol
 
-	acceptChan chan [2]routing.Addr
+	acceptChan chan routing.RouteDescriptor
```
evanlinjin, line 47: Don't bother working with `app`. We are removing it completely. Actually it will be good if we can remove it now. Please coordinate with @Darkren

pkg/router/gateway.go
```diff
@@ -0,0 +1,57 @@
...
+func NewGateway(router *router) *Gateway {
+	return &Gateway{
+		logger: logging.MustGetLogger("router-gateway"),
+		router: router,
+	}
+}
+
+func (r *Gateway) AddEdgeRules(rules routing.EdgeRules, ok *bool) error {
+	go func() {
+		r.router.accept <- rules
```
evanlinjin: We need to create a function such as: `router.IntroduceRules` or something similar. This is to allow for thread-safety when shutting down.

pkg/router/router.go
```diff
@@ -37,386 +33,367 @@ var log = logging.MustGetLogger("router")
...
-type Router struct {
-	Logger *logging.Logger
+type DialOptions struct {
+	MinForwardRts int
+	MaxForwardRts int
+	MinConsumeRts int
+	MaxConsumeRts int
+}
+
+var DefaultDialOptions = &DialOptions{
```
evanlinjin: I would prefer if this was a function instead: `func DefaultDialOptions() DialOptions`. This way the values become immutable.

pkg/router/router.go
```diff
@@ -37,386 +33,367 @@ var log = logging.MustGetLogger("router")
...
-func (r *Router) ServeApp(conn net.Conn, port routing.Port, appConf *app.Config) error {
-	fmt.Println("!!! [ServeApp] start !!!")
-
-	r.wg.Add(1)
-	defer r.wg.Done()
 
-	appProto := app.NewProtocol(conn)
-	if err := r.pm.Open(port, appProto); err != nil {
-		return err
+	req := routing.BidirectionalRoute{
```
evanlinjin, line 165: Nice name!

pkg/router/router.go
```diff
@@ -37,386 +33,367 @@ var log = logging.MustGetLogger("router")
...
-		if err := conn.Close(); err != nil {
-			r.Logger.WithError(err).Warn("Failed to close connection")
-		}
-	}
+// AcceptsRoutes should block until we receive an AddRules packet from SetupNode that contains ConsumeRule(s) or ForwardRule(s).
+// Then the following should happen:
+// - Save to routing.Table and internal RouteGroup map.
+// - Return the RoutingGroup.
+func (r *router) AcceptRoutes() (*RouteGroup, error) {
+	rules := <-r.accept
```
evanlinjin: We may need a `done` chan to go along with this. Probably need to give some thought on the close logic.

pkg/router/router.go
```diff
@@ -426,11 +403,44 @@ fetchRoutesAgain:
...
+		r.logger.Infof("Save new Routing Rule with ID %d %s", rule.KeyRouteID(), rule)
+	}
+
+	return nil
+}
+
+func (r *router) occupyRouteID(n uint8) ([]routing.RouteID, error) {
+	var ids = make([]routing.RouteID, n)
+	for i := range ids {
+		routeID, err := r.rt.ReserveKey()
```
evanlinjin: It might be better to change `(routing.Table).ReserveKey()` to `(routing.Table).ReserveKeys(n int)`.

pkg/router/router.go
evanlinjin: Or work some magic using `context.Context`.

pkg/router/routerclient/client.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	if err := c.rpc.Close(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *Client) AddEdgeRules(ctx context.Context, rules routing.EdgeRules) (bool, error) {
```
evanlinjin, line 55: If we use the deadline functions of the underlying `net.Conn` (`dmsg` that is), we won't need to use `context.Context`. Just some food for thought... Not sure which is best. But lets use `context.Context` for now.

pkg/setup/gateway.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (g *Gateway) DialRouteGroup(route routing.BidirectionalRoute, rules *routing.EdgeRules) (failure error) {
+	startTime := time.Now()
+	defer func() {
+		g.sn.metrics.Record(time.Since(startTime), failure != nil)
+	}()
+
+	g.logger.Infof("Received RPC DialRouteGroup request")
+
+	ctx := context.Background()
```
evanlinjin: We might need to have a context with timeout. Let's say 1 minute just to be safe? Is there a better way to do timeout?

pkg/setup/idreservoir.go
```diff
@@ -97,60 +102,51 @@ func (rm RulesMap) String() string {
...
-	if err != nil {
-		return nil, 0, 0, err
-	}
-	firstRevRID, lastRevRID, err := SaveForwardRules(rules, idc, ld.KeepAlive, ld.Reverse)
-	if err != nil {
-		return nil, 0, 0, err
-	}
+// - a map that relates a slice of routing rules to a given visor's public key.
+// - an error (if any).
+func (idr *idReservoir) GenerateRules(forward, reverse routing.Route) (forwardRules, consumeRules map[cipher.PubKey]routing.Rule, intermediaryRules RulesMap, err error) {
```
evanlinjin, line 110: I know that routes are working in a very different manner in m2, but there was a bug in m1 (fixed here: https://github.com/SkycoinProject/skywire-mainnet/commit/7eccea4421cfad767c21e0c4955f221f9f329106). Let me know if this was causing an issue, or if we are fine. Thanks!

pkg/setup/gateway.go
```diff
@@ -0,0 +1,121 @@
...
+
+func NewGateway(reqPK cipher.PubKey, sn *Node) *Gateway {
+	return &Gateway{
+		logger: logging.MustGetLogger("setup-gateway"),
+		reqPK:  reqPK,
+		sn:     sn,
+	}
+}
+
+func (g *Gateway) DialRouteGroup(route routing.BidirectionalRoute, rules *routing.EdgeRules) (failure error) {
```
evanlinjin: This looks clean. However, I would prefer if most of the logic is elsewhere, and the `Gateway` structure acts as just a *gateway*.

pkg/setup/setupclient/client.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/skycoin/skywire/pkg/routing"
+	"github.com/skycoin/skywire/pkg/snet"
+)
+
+const rpcName = "Gateway"
+
+type Client struct {
+	log   *logging.Logger
+	n     *snet.Network
+	nodes []cipher.PubKey
```
evanlinjin: We need to make clear (probably via comments) that this refers to setup node PKs.

evanlinjin (changes requested):
Very good so far. Please review my comments. Please don't worry about *Making it work as a whole* yet. Just get the components mentioned in the specs implemented. Let's look to merge this (it will be okay if it's not working), but get it compiling. Then let's rebase this to the SkycoinProject org before continuing further.

pkg/router/router.go
nkryuchkov: @evanlinjin thanks!

pkg/setup/idreservoir.go
nkryuchkov: @evanlinjin It's fixed in the new implementation.

pkg/setup/gateway.go
nkryuchkov: @evanlinjin 1 minute is a good idea, I'll think if there's a better way.

pkg/app/app.go
nkryuchkov: @Darkren will remove it in his PR

pkg/app2/rpc_gateway.go
```diff
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/pkg/errors"
```
evanlinjin, line 7: So have we decided against using `pkg/errors` for m2?

pkg/router/route_group.go
```diff
@@ -0,0 +1,218 @@
...
+
+// RouteGroup should implement 'io.ReadWriteCloser'.
+type RouteGroup struct {
+	mu sync.RWMutex
+
+	desc routing.RouteDescriptor // describes the route group
+	fwd  []routing.Rule          // forward rules (for writing)
+	rvs  []routing.Rule          // reverse rules (for reading)
+
+	// The following fields are used for writing:
```
evanlinjin, line 31: We need to reorganize these comments/fields.

pkg/router/route_group.go
```diff
@@ -0,0 +1,218 @@
...
+
+	return ioutil.BufRead(&r.readBuf, data, p)
+}
+
+// Write writes payload to a RouteGroup
+// For the first version, only the first ForwardRule (fwd[0]) is used for writing.
+func (r *RouteGroup) Write(p []byte) (n int, err error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
```
evanlinjin, line 94: Should we check if the RouteGroup is closed here, to avoid writing when the RouteGroup closes.

evanlinjin (changes requested):

evanlinjin:
I haven't said much. The code looks good. We just need to ensure it works.

pkg/app2/rpc_gateway.go
nkryuchkov: The `github.com/pkg/errors` is not related to the standard library. The new error features in 1.13 were added to the `errors` package. 

The `github.com/pkg/errors` was used in `app2` only, so the usage of `errors` and `github.com/pkg/errors` was mixed. I decided to make it consistent in this PR. 

I think we should go with the standard library `errors` and use more of its features but within another PR.

pkg/router/route_group.go
nkryuchkov: @evanlinjin You're right, thanks!

nkryuchkov:
Replaced by https://github.com/SkycoinProject/skywire-mainnet/pull/51

Issue #560, opened 2019-09-06
Migrate hypervisor to visor communication to dmsg


ivcosla:
In the communication between hypervisor and visor to use `dmsg` underneath instead of `tcp`

evanlinjin:
Note that we are handling `dmsg` communication via the `snet.Network` structure.

PR #563, opened 2019-09-06
dmsg hypervisor

ivcosla:
Working state, integration tests are working, but hypervisor tests need fix.


ivcosla:
Closes #560

pkg/hypervisor/config.go
```diff
@@ -35,12 +37,13 @@ func (hk *Key) UnmarshalText(text []byte) error {
...
-	EnableAuth bool            `json:"enable_auth"` // Whether to enable user management.
-	Cookies    CookieConfig    `json:"cookies"`     // Configures cookies (for session management).
-	Interfaces InterfaceConfig `json:"interfaces"`  // Configures exposed interfaces.
+	PK                 cipher.PubKey   `json:"public_key"`
+	SK                 cipher.SecKey   `json:"secret_key"`
+	DBPath             string          `json:"db_path"`             // Path to store database file.
+	EnableAuth         bool            `json:"enable_auth"`         // Whether to enable user management.
+	Cookies            CookieConfig    `json:"cookies"`             // Configures cookies (for session management).
+	Interfaces         InterfaceConfig `json:"interfaces"`          // Configures exposed interfaces.
+	MessagingDiscovery string          `json:"messaging_discovery"` // MessagingDiscovery address for dmsg usage
```
evanlinjin: DmsgDiscovery

evanlinjin (commented):
Good start.

pkg/hypervisor/hypervisor.go
```diff
@@ -32,42 +33,59 @@ var (
...
 type Node struct {
-	c     Config
-	nodes map[cipher.PubKey]appNodeConn // connected remote nodes.
-	users *UserManager
-	mu    *sync.RWMutex
+	c       Config
+	nodes   map[cipher.PubKey]appNodeConn // connected remote nodes.
+	users   *UserManager
+	mu      *sync.RWMutex
+	Network *snet.Network
```
evanlinjin: Maybe using `dmsg` directly will be less clunky? `snet` was only intended to be used with `visor`.

evanlinjin (changes requested):

evanlinjin:
Btw, will be good if you can provide steps of how this can be tested.

ivcosla:
This can be tested with `skywire-services` on `feature/dmsg-hypervisor` branch. Then:
 1) `make integration-build`
 2) `make integration-run-generic`

From now on you can test the hypervisor API, for example you can `curl localhost:8080/api/nodes`

pkg/hypervisor/hypervisor.go
ivcosla: I think I can just create a listener in hypervisor cmd app, then pass it down to `ServeRPC` and remove `Network` from `Node`

pkg/hypervisor/config.go
ivcosla: Yes, but note that I used the same naming that in `visor` config. I think it's time for us to change it in visor too.

ivcosla:
Note however that by removing `noise.RPCClientDialers` now there is no retrial in the `Dial` operation between `visor` and `hypervisor`. It could be added if deemed necessary in line https://github.com/skycoin/skywire/pull/563/files#diff-8b79394cd85c849e746408355107d3b6R241 using an `netutil.Retrier`.

cmd/hypervisor/commands/root.go
```diff
@@ -62,7 +62,27 @@ var rootCmd = &cobra.Command{
...
+			dmsgConf := snet.Config{
+				PubKey:     config.PK,
+				SecKey:     config.SK,
+				TpNetworks: []string{snet.DmsgType},
+
+				DmsgDiscAddr: config.DmsgDiscovery,
+				DmsgMinSrvs:  1,
+			}
+
+			network := snet.New(dmsgConf)
```
evanlinjin: We don't want the `snet` dependency in hypervisor. Use `dmsg` directly.

evanlinjin (changes requested):
> Note however that by removing `noise.RPCClientDialers` now there is no retrial in the `Dial` operation between `visor` and `hypervisor`. It could be added if deemed necessary in line https://github.com/skycoin/skywire/pull/563/files#diff-8b79394cd85c849e746408355107d3b6R241 using an `netutil.Retrier`.

Yes, we will need retry logic. Please implement.

evanlinjin:
> `make integration-build`

Where is the endpoint documentation kept? Are we still updating this Postman file?

https://github.com/skycoin/skywire/blob/mainnet-milestone2/cmd/hypervisor/hypervisor.postman_collection.json

ivcosla:
> > `make integration-build`
> 
> Where is the endpoint documentation kept? Are we still updating this Postman file?
> 
> https://github.com/skycoin/skywire/blob/mainnet-milestone2/cmd/hypervisor/hypervisor.postman_collection.json

I didn't update it, will do it as part of the requested changes.

pkg/visor/visor.go
```diff
@@ -247,12 +238,21 @@ func (node *Node) Start() error {
...
-	for _, dialer := range node.rpcDialers {
-		go func(dialer *noise.RPCClientDialer) {
-			if err := dialer.Run(rpcSvr, time.Second); err != nil {
-				node.logger.Errorf("Dialer exited with error: %v", err)
+	for _, hypervisor := range node.config.Hypervisors {
+		go func(hypervisor HypervisorConfig) {
+			var conn net.Conn
+			var err error
+
+			err = retrier.Do(func() error {
```
evanlinjin: The redial logic here is not the same as what we had previously in:
https://github.com/skycoin/dmsg/blob/70f4c32a994fe7fa62c36049e3ce46d0dc5c8c20/noise/net.go#L55-L77

Please fix.

evanlinjin (changes requested):

pkg/visor/visor.go
evanlinjin: To elaborate, if something breaks during `rpc` communication, we want to redial.

evanlinjin: To elaborate further, if hypervisor restarts, visor should be able to reconnect automatically to the hypervisor.

evanlinjin:
@ivcosla Just gave it a test run and it is working very well. Please fix the last request, thank you.

pkg/visor/rpc_client.go
```diff
@@ -213,6 +223,103 @@ func (rc *rpcClient) Loops() ([]LoopInfo, error) {
...
+// It exposes a RPC Server.
+// It will return if Close is called or crypto fails.
+func (d *RPCClientDialer) Run(srv *rpc.Server, retry time.Duration) error {
+	if ok := d.setDone(); !ok {
+		return ErrAlreadyServing
+	}
+	for {
+		if err := d.establishConn(); err != nil {
+			// Only return if not network error.
+			if err != dmsg.ErrRequestRejected {
```
ivcosla, line 254: Similar to `*noise.RPCClientDialer` check for `net.Error` type assertion here I think it would be useful to group all of `dmsg` network errors under a custom error type. Or at least those that we consider meaningful for checking outside the package.

